PR #27, opened 2025-04-24
Feature/add fact reasoner

jbrry:
### Summary :memo:
Moves the git subtree `fact_reasoner` of FactReasoner codebase into the `akd/tools/fact_reasoner`

### Details
Removed subtree from top-level of the repository
Re-applied `git subtree add --prefix akd/tools/fact_reasoner` command to move to the new location.


NISH1001:
@jbrry Are the json files in `examples/` required or they are just test files?

If they aren't required, can we remove them from git tree?

Thanks.

Issue #14, opened 2025-03-27
FactReasoner Integration


muthukumaranR:
### Is your feature request related to a problem?

```shell
Currently Factreasoner is not part of the framework. we need to add the codebase and figure out how it can be packaged (as a node) and reused across.
```

### Describe the solution you'd like

```shell
- add FactReasoner codebase
- Investigate feasibility to run within the langgraph environment.
- Discuss how best to use this, and implications.
```

### Describe alternatives you've considered

```shell

```

### Additional context

```shell

```

pascaleale:
Things to look at to start on this:
- Initial attempt to langgraph implementation to see how it would work
- Implementation of retrieval components of our pipeline, what it outputs, is that in the format needed for factuality checking
- FactReasoner includes LLMs in the pipeline, review how it's run now and if we need alternatives
- Where we want to run FactReasoner, which agent should have it?
- Evaluation, how we evaluate performances, we may need a human evaluation pilot here

NISH1001:
TODO:
- Run experiments and benchmarks for specific cases
- Technical Deep Dive

NISH1001:
cc: @jbrry Initial implementation with test sample query

![Image](https://github.com/user-attachments/assets/c0ba1ff1-af8d-42b6-be56-06941d7771cd)

PR #30, opened 2025-05-02
Add Vector Database to LitAgent

jbrry:
### Summary :memo:
For greater forensics, and to provide contexts to FactReasoner, a vector database would be useful for storing the retrieved contents.

This PR adds a vector database tool to the `LitAgent`, and indexes those results so they can be queried later.

### Details
- Add VectorDB tool with associated schemas for input/output for indexing and querying
- Added section in toml config file, and added a config for the tool
- Added the tool to LitAgent
- Updated LitAgent `arun` method to index documents as they are returned from the search tool.

### Checks
- Indirectly related to #14


akd/agents/litsearch.py
```diff
@@ -50,9 +62,10 @@ def __init__(
         query_agent: QueryAgent,
         extraction_agent: EstimationExtractionAgent,
         search_tool: SearxNGSearchTool,
+        vector_search_tool: VectorDBSearchTool,
```
NISH1001: Let's make the vetordb search optional for time being. Use only if provided.

akd/agents/litsearch.py
```diff
@@ -121,6 +135,39 @@ async def arun(self, params: LitAgentInputSchema) -> LitAgentOutputSchema:
...
+            langchain_docs.append(
+                Document(page_content=content.result, metadata=sanitized_metadata),
+            )
+
+        logger.info(f"Indexing {len(langchain_docs)} documents into Vector DB...")
+        await self.vector_search_tool.arun_index(
+            VectorDBIndexInputSchema(
+                documents=langchain_docs,
+            ),
+        )
```
NISH1001: Add a check if the vectordb tool is provided, only index..else run normally without indexing

NISH1001 (changes requested):

NISH1001:
@jbrry I was thinking maybe the lit search itself should be lightweight. The lit agent is solely used for structured extraction, so I am not sure if using a vector makes sense. Can we have the vector database for now as an independent tool? The flow is still not strong, IMO, on where the vector fits. Maybe it will be used for the Lit review agent.

Also, the litagent we have right now, IMO, might have to be renamed as something else. What do you think.

Nevertheless, I have made comments for review. 

Thanks for the PR. Let's keep the db as an independent component for the time being, and we can choose to integrate it into other components after we figure out how/where to integrate (could be at both on storm and here as well).

NISH1001 (approved):

akd/agents/litsearch.py
```diff
@@ -131,6 +186,20 @@ async def arun(self, params: LitAgentInputSchema) -> LitAgentOutputSchema:
...
+            sample_result = results[0]
+            print("Outputs of FR")
+            fr_output = await self.fact_reasoner_tool.arun(
+                FactReasonerInputSchema(
+                    response=sample_result,
+                ),
+            )
+
+            print(fr_output)
+            breakpoint()
```
NISH1001: Remove breakpoint. For print, using `logger.debug` which is imported using `from loguru import logger`

NISH1001:
Let's change the PR to develop now:
https://github.com/NASA-IMPACT/accelerated-discovery/tree/develop

PR #44, opened 2025-05-31
Notebook/deepeval dag evaluation

leothomas:
### Summary :memo:
This PR adds a notebook with an initial implementation of the DeepEval DAG evaluation metric. For now the graph is very simple and just does 2 content relevancy checks - I will continue to add more. 

### Details
I've added 2 important elements: 
- I've create a `CustomTaskNode` class that inherits from the `TaskNode` class. The `TaskNode` is intended to perform certain tasks before going to evaluation (eg: extract titles or URLs from the agent output), using LLM prompt instructions. However, since our results output is Pydantic model, rather than trusting the LLM to extract the correct elements from the JSON dump of that model, I figured it would be much more efficient to simply load the dictionary and fetch the correct keys. So I've create the `CustomTaskNode` that executes an arbitrary function on the LLMOutput before passing it to the evaluation condition. 
- I've added a very simple `graphviz` based visualization step to represent the graph. Note that the `graphviz` package needs to be instance (`brew install graphviz`) on Mac

### Usage: 

The goal of the `LLMEvaluator` is to provide an easy way to access LLM-as-Judge functionality that can be repurposed in a variety of places within the planning/execution phases of the agent. In theory, this can also be used to bounce the execution back to the planning phase, if the evaluation of intermediate or final outputs finds them un-satisfactory.

The `LLMEvaluator` uses a locally running model (`IBM: grantie3-dense:8b`) but can be easily updated to use any models running locally on Ollama using the following config: 
`file: .deepeval/.deepeval`
```
{
    "USE_AZURE_OPENAI": "NO",
    "LOCAL_MODEL_NAME": "granite3-dense:8b",
    "LOCAL_MODEL_BASE_URL": "http://localhost:11434",
    "LOCAL_MODEL_API_KEY": "ollama",
    "LOCAL_MODEL_FORMAT": "json",
    "USE_LOCAL_MODEL": "YES"
}
```

or to use any OpenAI models by setting the `USE_LOCAL_MODEL` key to `NO` in the settings above. In this case the 
`OPENAI_API_KEY` value needs to be set. When using the OpenAI` models, `OpenAI: gpt-4.1` is used by default

The `LLMEvaluator` class has two main usage modes: 

1. **Default evaluation metrics** If not instantiated with any specific evaluation metrics, the evaluator will default to the preset evaluation metrics, which are a series of DAG (Directed Acyclic Graphs) evaluators to assess the search results in 5 different dimensions (Usefulness, Faithfulness, Completeness, Accuracy, Timeliness), with the evaluation steps in each DAG becoming more and more restrictive. At the end, the result is given a score 0-10. The `threshold` parameter is used to define the threshold score for which the evaluation is considered a success. Once the evaluation has been run using `evaluator._arun()` the attribute `evaluation.score` will contain the averaged score for the evaluation along each dimension. The boolean attribute `evaluator.success` will indicate wether this average score is above or below the threshold value. 

```python
# Instantiate an evaluator class with the default eval metrics and threshold (defined using the config object): 
evaluator_config = LLMEvaluatorConfig(threshold=5.0)
evaluator = LLMEvaluator(config=evaluator_config, debug=self.debug)

# Create the evaluation input for a search result: 
evaluator_input = LLMEvaluatorInputSchema(
    input=params.query, output=json.dumps(content.result.model_dump())
)

# Run the evaluation
evaluation = await evaluator._arun(evaluator_input)

# Inspect the evaluation result for a search result and collect to return to user only if it has passed the evaluation
if not evaluation.success:
    logger.warning(
        f"Low evaluation score ({evaluation.score}) for content from {content.source}, skipping."
    )
else: 
    _results.append(content)
```
2. **Custom evaluation metrics:** To evaluate using a custom (non-default) metric, users can pass one or more dictionaries with evaluation criteria to the LLMEvaluator config object. This will instantiate one `GEval` evaluation metric  (as opposed to a DAG metric above) per criteria. GEval stands for `General Evaluation` and is the most general purpose evaluator that can accept any user defined criteria. It's worth noting that the DAG evaluators are essentially a series of `GEval` nodes strung together with custom scoring at each end node and an easy way to define different evaluation paths. The score of a GEval metric evaluation is in the range [0-1] - so the threshold should be set accordingly.

```python 
# Define custom evaluation criteria to evaluate how well the litsearch agent surfaced the relevant information 
extraction_structure_metric = {
    "name": "Extraction Structure Evaluation",
    "criteria": "Does the content includes a title and an abstract?",
}
extraction_quality_metric = {
    "name": "Extraction Evaluation",
    "criteria": "Evaluate wether or not the answer correctly extracts and summarizes the content, in a way that is relevant to the input query",
}

# Define an evaluator with the custom metrics, and again, a threshold value for success
extraction_evaluator_config = LLMEvaluatorConfig(
    custom_metrics=[extraction_quality_metric, extraction_structure_metric],
    threshold=0.5,
)

# As with the above example, create an evaluator input and run the evaluation: 

evaluator_input = LLMEvaluatorInputSchema(
    input=params.query, output=json.dumps(content.result.model_dump())
)

# Run the evaluation
extraction_evaluation = await evaluator._arun(evaluator_input)

# Inspect the evaluation result for a search result and collect to return to user only if it has passed the evaluation
if not extraction_evaluation.success:
    logger.warning(
        f"Low evaluation score ({evaluation.score}) for content from {content.source}, skipping."
    )
else: 
    _results.append(content)
```

### Next steps: 

@TigranTigranTigran and I have chatted about merging this evaluation work with the Granite Guardian work he has been doing to identify a range of other evaluation criteria relating to bias, toxicity, etc. 

We would like to expand the default capabilities of the `LLMEvaluator` class such that if it's passed only a `input: str`  parameter, it would know to evaluate a user's input query for any inappropriate or illegal questions, prompt injection, jailbreaking, etc. If passed a set of search results it would perform the same default evaluation as described above. If provided search results and an output, it would evaluate the output in the context the retrieved results. Each of these evaluations will be able to rely on the pre-defined metrics or custom user defined metrics.  

### Checks
- [ ] Closed #798
- [ ] Tested Changes
- [ ] Stakeholder Approval


NISH1001:
@leothomas Can we also have markdown-version of the evaluation. A summary or screenshot of expected results? Thanks.

.deepeval/.deepeval
```diff
@@ -0,0 +1,8 @@
+{
+    "USE_AZURE_OPENAI": "NO",
+    "LOCAL_MODEL_NAME": "granite3-dense:8b",
+    "LOCAL_MODEL_BASE_URL": "http://localhost:11434",
+    "LOCAL_MODEL_API_KEY": "ollama",
+    "LOCAL_MODEL_FORMAT": "json",
+    "USE_LOCAL_MODEL": "YES"
+}
```
NISH1001, line 8: We might have to consolidate the config to the codebase config at `configs`

.deepeval/.deepeval_telemetry.txt
```diff
@@ -0,0 +1,4 @@
+DEEPEVAL_ID=980c627b-422e-43d8-b6ce-78e5cff052eb
+DEEPEVAL_STATUS=old
+DEEPEVAL_LAST_FEATURE=evaluation
+DEEPEVAL_EVALUATION_STATUS=old
```
NISH1001: Is this a temporary artifact?

searxng/settings.yml
```diff
@@ -0,0 +1,2766 @@
+general:
```
NISH1001: Wondering if we really need searxng config in akd. WE're assuming searxng will be run as a separate instance somewhere.

pascaleale:
@leothomas you could start including these nodes:

They are taken from this [taxonomy](https://github.com/IBM/risk-atlas-nexus/blob/main/src/risk_atlas_nexus/data/knowledge_graph/granite_guardian_dimensions.yaml#L23) . IN the yaml you will find this structure for each risk:
```
- id: <>
  name: <>
  description: <>

```
I have pre-selected some of the risks and used name and description in the file to prompt LLama 405b with the following instruction:

```
Can you generate a criteria for an LLM as a judge pipeline, aimed at evaluating the risk described by the following name and description:

name:<>
description: <>

The criteria needs to be in a question format. Stay faithful to the details provided
```

The nodes are divided following the task you are evaluating. If there are other steps you want to evaluate that I haven't included please let me know 

```
#in case of output from retriever

check_structure_node = BinaryJudgementNode(
      criteria="Does the content includes a title and an abstract?", 
      children= [
          VerdictNode(verdict = False, score=0),
          VerdictNode(verdict = True, score=5, child=usefullness_of_content_node)
      ]
)

    
usefullness_of_content_node = BinaryJudgementNode(
    criteria="Does the content of the paper/set of papers directly address the query's main topic?",
        VerdictNode(verdict = False, score=0),
        VerdictNode(verdict = True, score=5, child= novel_insights_node)
    ]
)

novel_insights_node = BinaryJudgementNode(
      criteria="Does the content offer information, arguments, or perspectives that are likely new or non-obvious to the user?", 
      children= [
          VerdictNode(verdict = False, score=0),
          VerdictNode(verdict = True, score=5, child=None)
      ]
)

#in case of a summarization step

Summarization_importance_node = BinaryJudgementNode(
    criteria="Does the content include important information from the original text?",
        VerdictNode(verdict = False, score=0),
        VerdictNode(verdict = True, score=5, child= Summarization_faithfull_node)
    ]
)

Summarization_faithfull_node = BinaryJudgementNode(
    criteria="Is the content faithful to the original text?",
        VerdictNode(verdict = False, score=0),
        VerdictNode(verdict = True, score=5, child= None)
    ]
)

#in case of general output from an intermediate step with an LLM, considers a context and a query

LLM_generatino_Groundedness_node = BinaryJudgementNode(
    criteria="Is the content supported by the context provided?",
        VerdictNode(verdict = False, score=0),
        VerdictNode(verdict = True, score=5, child= LLM_generation_Answer_relevancy_node)
    ]
)

LLM_generation_Answer_relevancy_node = BinaryJudgementNode(
    criteria="Does the content directly and appropriately address the user’s query, staying focused on the intent and scope of the question without introducing unrelated or off-topic information?",
        VerdictNode(verdict = False, score=0),
        VerdictNode(verdict = True, score=5, child= None)
    ]
)
```

akd/tools/evaluator/base_evaluator.py
```diff
@@ -0,0 +1,95 @@
...
+    usefulness,
+    faithfulness,
+    completeness,
+    accuracy,
+    timeliness,
+)
+
+
+# from deepeval import Evaluator, SomeMetric
+class LLMEvaluatorInput(BaseToolConfig):
```
NISH1001: This should be inherited from `InputSchema` which is at `akd._base.InputSchema`

akd/tools/evaluator/base_evaluator.py
```diff
@@ -0,0 +1,95 @@
...
+    input: str
+    output: str
+    # TODO: implement reference output checking
+    reference: Optional[str] = None
+
+
+class SingleEvaluationOutput(BaseToolConfig):
```
NISH1001: Same: Use `OutputSchema`

`from akd._base import OutputSchema`

akd/tools/evaluator/base_evaluator.py
NISH1001: Also have the naming convetion as `XYZSchema` for consistency

akd/tools/evaluator/base_evaluator.py
```diff
@@ -0,0 +1,95 @@
...
+    score: float
+    reason: str
+    metric: str
+
+
+class LLMEvaluatorOutput(BaseToolConfig):
```
NISH1001: Use `OutputSchema`

akd/tools/evaluator/base_evaluator.py
```diff
@@ -0,0 +1,95 @@
...
+
+    score: float
+    evaluations: List[SingleEvaluationOutput]
+    success: bool
+
+    # Add any additional fields needed for the output schema
+
+
+class LLMEvaluator(BaseTool):
```
NISH1001, line 74: Have a config as well like
```python

class LLMEvaluatorConfig(BaseToolConfig)
...
..


class LLMEvaluator(BaseTool):
  input_schema = <...>
  output_schema = <...>
  config_schema = LLMEvaluatorConfig
```

akd/tools/evaluator/base_evaluator.py
```diff
@@ -0,0 +1,95 @@
...
+
+    input_schema = LLMEvaluatorInput
+    output_schema = LLMEvaluatorOutput
+
```
NISH1001, line 79: needs `config_schema` as well for better tooling.

akd/tools/evaluator/base_evaluator.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Modular evaluator for LLM outputs using configurable criteria and deepeval metrics.
+    Loads criteria from a JSON file and exposes an evaluate method for integration.
+    """
+
+    def __init__(self, custom_metrics: List[Dict] = [], threshold: float = 0.5):
```
NISH1001: bass this custom metrics through config `LLMEvaluatorConfig` should have these parameters.

Check any tools at `akd.tools` for understanding the structure better.

The reason is `BaseTool` will handle everything then.

notebooks/end-to-end-debug.ipynb
```diff
@@ -1074,7 +1074,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.0"
+   "version": "3.12.10"
```
NISH1001: Let's not push notebook changes please unless it's a major change. I'd rather push notebooks later.

NISH1001 (changes requested):

searxng/settings.yml
leothomas: The settings file was autogenerated when I ran SEARXNG - I can remove it and add it to the .gitignore

leothomas:
Hey folks! Just pushed a few updates to better adhere to the code conventions (re: tooling/configurations/input and output schema naming etc)

@pascaleale I added the `check_structure_node` criteria to a GEval step which runs prior to the content evaluation. The present content evaluation metric already include evaluations similar to that in the `usefulness_of_content` and `novel_insights` nodes you shared. 

For last 4 nodes you suggested: `Summarization_importance_node, Summarization_faithfull_node , LLM_generatino_Groundedness_node , LLM_generation_Answer_relevancy_node `, they all require the context to be passed to the evaluation class, in addition to the model output. I think this would be super important functionality to add to the evaluation class, but I'd prefer to open a separate issue for that so we can close this PR. Thoughts?

.deepeval/.deepeval
leothomas: This file is written by DeepEval - I'm not seeing a very clear way to override the location of the config (or even the use of an external config file, rather than something like a Pydantic BaseSettings config)

leothomas: Ref: https://github.com/confident-ai/deepeval/blob/aced5e9280b790309bc5af3a7329502709ed1ce9/deepeval/key_handler.py

leothomas: Ref: https://github.com/confident-ai/deepeval/blob/aced5e9280b790309bc5af3a7329502709ed1ce9/deepeval/constants.py

akd/_base.py
```diff
@@ -166,7 +166,15 @@ def _post_init(self) -> None:
...
         if self.config is None:
             return
-        for attr, value in self.config.model_dump().items():
+
+        # Convert BaseConfig to dict if it's a BaseModel
+        config = (
+            self.config.model_dump()
+            if isinstance(self.config, BaseConfig)
+            else self.config
+        )
```
NISH1001: Let's remove this logic. We're guaranteed to have an object of type BaseConfig. That's the whole point of the new structure. There's no point in doing it because there's already a `from_dict` method that can create object using dictionary. The base is enforcing pydantic model strictly for config. If not, the downstream implementation has problem, not the base.

NISH1001: Also, try avoiding modifying the base classes because it'll need discussion around it. PR should be only concerned about creating new components. If the base has a bug, we do bugfixes separately in a new PR.

akd/agents/litsearch.py
```diff
@@ -142,14 +156,63 @@ async def _arun(
...
+            evaluation = await evaluator._arun(evaluator_input)
+
+            if not evaluation.success:
+                logger.warning(
+                    f"Low evaluation score ({evaluation.score}) for content from {content.source}, skipping."
+                )
+                continue
+            _results.append(content)
+
+        return LitAgentOutputSchema(results=_results)
```
NISH1001, line 220: Also note: LitAgent might need more restructuring (Which is on me than this PR). Since there are some changes to design we have thought. But this looks good for time being. However, evaluation I think shouldn't happen within agent/tool level. It should be outside and should work on output of these tools and agents, not modifying them in-place ad-hoc. what do you think?

akd/tools/evaluator/base_evaluator.py
```diff
@@ -0,0 +1,131 @@
...
+                        LLMTestCaseParams.ACTUAL_OUTPUT,
+                    ],
+                )
+                for m in config.custom_metrics
+            ]
+            if config.custom_metrics
+            else config.default_metrics
+        )
+
+        self.threshold = config.threshold
```
NISH1001: FYI (just a little nitpick):
BTW, this logic can be put in `_post_init(...)` method directly and we can avoid overriding the constructor unless significant parameter change happens within constructor arguments.

Eg;

```python

class LLMEvaluator(...):
  ...
  ...
  def _post_init(self) -> None:
    self.metrics = (
                [
                    GEval(
                        **m,
                        evaluation_params=[
                            # TODO: add option to use reference output
                            LLMTestCaseParams.INPUT,
                            LLMTestCaseParams.ACTUAL_OUTPUT,
                        ],
                    )
                    for m in config.custom_metrics
                ]
                if config.custom_metrics
                else config.default_metrics
            )
```
should work. in the `akd._base.AbstractBase` the constructor automatically calls the post init directly.

This is something that provides more flexibility in doing more init/setup level control without having to repeat the same constructor pattern.

akd/tools/evaluator/base_evaluator.py
```diff
@@ -0,0 +1,131 @@
...
+        return LLMEvaluatorOutputSchema(
+            score=avg_score,
+            success=avg_score >= self.threshold,
+            evaluations=[
+                SingleEvaluationOutputSchema(
+                    score=metric.score, reason=metric.reason, metric=metric.name
+                )
+                for metric in self.metrics
+            ],
+        )
```
NISH1001, line 136: nice...thanks

searxng/uwsgi.ini.new
```diff
@@ -0,0 +1,55 @@
...
+add-header = Connection: close
+
+# Follow SIGTERM convention
+# See https://github.com/searxng/searxng/issues/3427
+die-on-term
+
+# uwsgi serves the static files
+static-map = /static=/usr/local/searxng/searx/static
+static-gzip-all = True
+offload-threads = %k
```
NISH1001: As I think about it now, maybe having this at project config folder makes sense?
Like `<root>/config/searxng/`? What do you think?

NISH1001 (changes requested):

akd/_base.py
NISH1001: Let me know if any further clarification is required. FOr time being, let's revert this assuming everything is guaranteed to be a pedantic object.

pascaleale:
@leothomas regarding your comment on those nodes I added essentially these would be included in the modification of the structure of the node Tigran is working on as he is integrating GG models and two of them cover exactly the cases you mention, so should be included there.

NISH1001:
@leothomas could you also please add a `usage` section in this PR. Some examples are #66 #67 

Thanks

akd/_base.py
leothomas: Sounds good! I was running into an issue with the config not being a Pydantic object - but I completely agree with you that the correct solution is to make sure that `config` is well formed, rather than converting it elsewhere!

akd/agents/litsearch.py
leothomas: I'm of a split mind on this - we had discussed using the Evaluation class to both evaluate the ability for the lit search's agent to correctly extract and summarize the search output (eg: do the outputs all contain the required keys: abstract, title, url, etc) and the search's ability to surface good quality results (from a reputable journal, relevant to the user's query, provides novel insights, etc). The idea is to only perform the second evaluation once the first has passed, which would entail running it within the agent. 

However, I agree with you that to harness the power of the agent + tooling structure we should have the orchestrator/planner node know to call the evaluation protocol after the search tool and then possibly restart the search with updated parameters, if the search output fails the evaluation screen.

Should we keep the evaluation within the search agent for now, and the revisit after thursday's demo?

searxng/uwsgi.ini.new
leothomas: Sorry - this file was also autogenerated when running searxng locally - I meant to add it to the .gitignore file. Since it's autogenerated and I haven't made any updates to it I think we can safely remove it from source control

NISH1001:
Closing this as stale for time being.

PR #48, opened 2025-06-18
Refactor atomic-agents to use LangGraph

movinam:
### Summary
This PR works on replacing atomic agents with its Pydantic and Langchain equivalent. This is because most of the atomic agents code is just a wrapper around Pydantic, which we can directly implement instead of using it through atomic agents remove our dependency on the package

### Details
1. The PR replaces BaseIOSchema with BaseModel since BaseIOSchema implement BaseModel. This way, we can implement our own Schema's without going through the checks imposed by BaseIOSchema.
2. It also replaces AtomicBaseAgent with BaseModel. This is to remove the need certain functionalities such as SystemPromptGenerator and AgentMemory, which can be implemented from node to node without wrapping the code around it.
3. BaseAgentConfig is replaced by ConfigDict. This allows client to implemented without using the instructor library.

### TODO
1. The SystemPromptGenerator still has to be replaced with ChatPromptTemplate or its equivalent.
2. Since LangGraph implements tooling and asynchronous calling, we have to address the need for AsyncRunMixin and LangchainToolMixin.
3. Run tests over the existing tools and nodes to check if they work as expected:

### DIFF
--- a/akd/agents/factory.py
+++ b/akd/agents/factory.py
@@ -1,7 +1,6 @@
 from typing import List, Optional, Union
 
-import instructor
-import openai
+from langchain_openai import ChatOpenAI
 from pydantic import create_model, BaseModel, ConfigDict
 
 from akd.configs.project import CONFIG
@@ -15,8 +14,9 @@ from .query import QueryAgent
 
 def create_intent_agent(config: Optional[ConfigDict] = None) -> IntentAgent:
     config = config or ConfigDict(
-        client=instructor.from_openai(
-            openai.AsyncOpenAI(api_key=CONFIG.model_config_settings.api_keys.openai),
+        client=ChatOpenAI(
+                    api_key=CONFIG.model_config_settings.api_keys.openai,
+
         ),
         model=CONFIG.model_config_settings.model_name,
         system_prompt_generator=SystemPromptGenerator(
@@ -50,8 +50,9 @@ def create_extraction_agent(
     )
     return BaseModel(
         ConfigDict(
-            client=instructor.from_openai(
-                openai.OpenAI(api_key=CONFIG.model_config_settings.api_keys.openai),
+            client=ChatOpenAI(
+                    api_key=CONFIG.model_config_settings.api_keys.openai,
+
             ),
             model=CONFIG.model_config_settings.model_name,
             # TODO: Replace with LangGraph equivalent
@@ -91,9 +92,10 @@ def create_extraction_agent(
 
 def create_query_agent(config: Optional[ConfigDict] = None) -> QueryAgent:
     config = config or ConfigDict(
-        client=instructor.from_openai(
-            openai.AsyncOpenAI(api_key=CONFIG.model_config_settings.api_keys.openai),
-        ),
+        client=ChatOpenAI(
+                    api_key=CONFIG.model_config_settings.api_keys.openai,
+
+            ),
         model=CONFIG.model_config_settings.model_name,
         # TODO: Replace with LangGraph equivalent
         system_prompt_generator=SystemPromptGenerator(
--- a/akd/agents/factory.py
+++ b/akd/agents/factory.py
@@ -1,11 +1,12 @@
 from typing import List, Optional, Union
 
 from langchain_openai import ChatOpenAI
+from langchain.prompts import ChatPromptTemplate, MessagesPlaceholder
 from pydantic import create_model, BaseModel, ConfigDict
 
 from akd.configs.project import CONFIG
+from akd.configs.prompts import INTENT_SYSTEM_PROMPT, EXTRACTION_SYSTEM_PROMPT, QUERY_SYSTEM_PROMPT
 from akd.structures import ExtractionSchema, SingleEstimation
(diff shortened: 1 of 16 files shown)


akd/agents/factory.py
```diff
@@ -2,20 +2,19 @@
...
 from .intents import IntentAgent
 from .query import QueryAgent
 
 
-def create_intent_agent(config: Optional[BaseAgentConfig] = None) -> IntentAgent:
-    config = config or BaseAgentConfig(
+def create_intent_agent(config: Optional[ConfigDict] = None) -> IntentAgent:
+    config = config or ConfigDict(
         client=instructor.from_openai(
             openai.AsyncOpenAI(api_key=CONFIG.model_config_settings.api_keys.openai),
```
jbrry: I'm wondering can we make this:
`client=ChatOpenAI(api_key=CONFIG.model_config_settings.api_keys.openai,)`

Do we still want to use `instructor`?

akd/agents/factory.py
```diff
@@ -46,15 +45,16 @@ def create_extraction_agent(
...
         answer=(extraction_output_schema, "Answer from the extraction"),
-        __base__=BaseIOSchema,
+        __base__=BaseModel,
         __doc__="Extracted information from the research",
     )
-    return BaseAgent(
-        BaseAgentConfig(
+    return BaseModel(
+        ConfigDict(
             client=instructor.from_openai(
```
jbrry: same as above re `instructor`

akd/agents/factory.py
```diff
@@ -89,12 +89,13 @@ def create_extraction_agent(
     )
 
 
-def create_query_agent(config: Optional[BaseAgentConfig] = None) -> QueryAgent:
-    config = config or BaseAgentConfig(
+def create_query_agent(config: Optional[ConfigDict] = None) -> QueryAgent:
+    config = config or ConfigDict(
         client=instructor.from_openai(
```
jbrry: same comment re instructor

akd/agents/factory.py
jbrry: I see in `scripts/run_lit_agent.py` below `config` is actually `ChatOpenAI` so the `or ConfigDict` will be different if it is not passed.

movinam: Missed these instances. They have been removed in the latest commit. Thanks James!

akd/tools/fact_reasoner_tool.py
```diff
@@ -15,10 +15,10 @@
 from akd.tools.fact_reasoner.fm_factual.nli_extractor import NLIExtractorOld
 from akd.tools.fact_reasoner.fm_factual.query_builder import QueryBuilder
 
-from ._base import BaseIOSchema, BaseTool, BaseToolConfig
+from ._base import BaseTool
 
 
-class FactReasonerToolConfig(BaseToolConfig):
+class FactReasonerToolConfig(ConfigDict):
```
jbrry, line 21: I am not sure if we can subclass ConfigDict, I tried and got some [class not fully defined errors](https://docs.pydantic.dev/latest/errors/usage_errors/#class-not-fully-defined)

For ConfigDict, I am normally seeing this pattern, the class sub-classes BaseModel and config dict is a parameter: https://docs.pydantic.dev/2.0/usage/model_config/

We might just be able to stick with BaseModel for now

movinam: I'm not able to test FactReasoner on my end. Can you test it with BaseModel and push a commit to the branch if it works please? You can also try the following:

```
from pydantic_settings import BaseSettings, SettingsConfigDict

class FactReasonerToolConfig(BaseSettings):
    model_config = SettingsConfigDict()
    gen_model: str
    context_retriever_service_type: str
    cache_dir: str
    collection_name: str
    k: int
    merlin_path: str
```

NISH1001: BaseToolConfig is just a pydantic type. ConfigDict is not for this purpose. We can have it as a type with some attributes auto-generated..like id, name etc...

```python

class BaseToolConfig(BaseModel):
  name: Optional[str] = Field(None,...)
...
```

NISH1001:
Based on some thoughts revolving around it, let's do this:
- Have `BaseAgent` completely as an abstract base class 

```python

class InputSchema(BaseModel):
  ....

class OutpuSchema(BaseModel):
  ...


class BaseAgent[InputSchema: BaseModel, OutputSchema: BaseModel](ABC, some mixins we already have):

  def history(self, *args, **kwargs) -> List[Any]:

    raise NotIMplementedError()

  async def get_response_async(...):
    raise NotImplementedError()

  # user has to implement this
  @abstractmethod
  async def arun(self, param: InputSchema, **kwargs):
    raise NotImplementedError()
  ....
  ...


# Impl

# Movina's langchain-based impl can go here
class LangBaseAgent(BaseAgent):
 <implements its own thing>


# 

# whatever there was previously here or just probably removing this in favor of pydnatic-ai might be btter
class AtomicBaseAgent(BaseAgent):
  ...


class PydanticAIBaseAgent(BaseAgent):
  ...
```

NISH1001:
We can close this in favor or #50

PR #61, opened 2025-07-03
Feature/granite guardian tool

TigranTigranTigran:
### Summary :memo:
A tool that calls a Granite Guardian model to detect risks. The input schema allows detection in single- or multi-turn "conversations". It also allows for the detection of risks in a SearxNGSearchToolOutputSchema object, which triggers multi-turn risk detection (where the "user" message is the query and the "assistant" message is the content of the search item).

A "tests" directory has been added to the root containing a unit test for the new tool.

### Details
1. The tool calls Granite Guardian in Ollama, first setting one of its risk definitions
2. The unit test creates synthetic search results (instances of SearchResultItem) and places them in an instance of SearxNGSearchToolOutputSchema in order to emulate the output of the search tool. It then tests for risk within "answer relevance" by submitting to Granite Guardian multi-turn conversations wherein the user messages are the search queries and the assistant messages are the "content" fields of the search items.

### Usage

```python
from akd.tools.granite_guardian_tool import (
    GraniteGuardianInputSchema,
    GraniteGuardianOutputSchema,
    GraniteGuardianTool,
    GraniteGuardianToolConfig,
    RiskDefinition,
    GuardianModelID,
    OllamaType
)

    SearxNGSearchToolInputSchema,
    SearxNGSearchToolOutputSchema,
    SearxNGSearchTool
)
```

```python
search_tool = SearxNGSearchTool()
search_results = await search_tool.arun(
    SearxNGSearchToolInputSchema(queries=["landslide in nepal last year"])
)

search_results
```

```python
config = GraniteGuardianToolConfig(
        model=GuardianModelID.GUARDIAN_2B,
        default_risk_type=RiskDefinition.ANSWER_RELEVANCE,
        ollama_type=OllamaType.CHAT,
        )

tool = GraniteGuardianTool(config=config, debug=True)
search_results = <instance of SearxNGSearchToolOutputSchema or SearchToolOutputSchema>

input_schema = GraniteGuardianInputSchema(
        search_results=search_results,
        risk_type=RiskDefinition.ANSWER_RELEVANCE,
    )

output = await tool.arun(input_schema)
```

### Checks
- [ ] Closed #798
- [x] Tested Changes
- [ ] Stakeholder Approval


NISH1001:
@TigranTigranTigran Thanks for this. I Will review this once #62 is merged.

akd/tools/granite_guardian_tool.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class GraniteGuardianInputSchema(InputSchema):
+    """
+    Input schema for the Granite Guardian tool.
+    """
+
+    query: Optional[str] = Field(None, description="Query text to evaluate.")
+    response: Optional[str] = Field(None, description="Assistant/model response.")
+    search_results: Optional[SearxNGSearchToolOutputSchema] = Field(
```
NISH1001: I think we could do `akd.tools.search.SearchToolOutputSchema` so that it'll be more polymorphic?

akd/tools/granite_guardian_tool.py
```diff
@@ -0,0 +1,180 @@
...
+        "social_bias",
+        "profanity",
+        "sexual_content",
+        "unethical_behavior",
+        "violence",
+        "jailbreak",
+        "groundedness",
+        "relevance",
+        "answer_relevance",
+    ]
```
NISH1001: Can we have this as a list of enums?

NISH1001: Also, let's rename `sexual_content` to `NSFW`?

akd/tools/granite_guardian_tool.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+
+
+class GraniteGuardianToolConfig(BaseToolConfig):
+    """
+    Configuration for Granite Guardian Tool.
+    """
+
+    model: str = "granite3-guardian"
+    default_risk_type: str = "answer_relevance"
```
NISH1001: Add pydantic way `Field(...)`, so that we can add description and default values/validation.

NISH1001: `    default_risk_type: RiskDefinition = Field(default=RiskDefinition.ANSWER_RELEVANCE)`

akd/tools/granite_guardian_tool.py
NISH1001:
```python
from enum import Enum

class RiskDefinition(Enum):
    """
    Enumeration of possible risk definitions.
    """
    HARM = "harm"
    SOCIAL_BIAS = "social_bias"
    PROFANITY = "profanity"
    NSFW = "nsfw"
    UNETHICAL_BEHAVIOR = "unethical_behavior"
    VIOLENCE = "violence"
    JAILBREAK = "jailbreak"
    GROUNDEDNESS = "groundedness"
    RELEVANCE = "relevance"
    ANSWER_RELEVANCE = "answer_relevance"
```

akd/tools/granite_guardian_tool.py
```diff
@@ -0,0 +1,180 @@
...
+
+    async def _arun(
+        self,
+        params: GraniteGuardianInputSchema,
+    ) -> GraniteGuardianOutputSchema:
+        risk_type = params.risk_type or self.default_risk_type
+        if risk_type not in self.allowed_risk_definitions:
+            raise ValueError(
+                f"Risk check type '{risk_type}' is not supported. Allowed values are: {self.allowed_risk_definitions}",
+            )
```
NISH1001: Using pydantic field (see previous review), I don't think we'll need this check as pydantic will automatically validate the enum

akd/tools/granite_guardian_tool.py
```diff
@@ -0,0 +1,180 @@
...
+        debug: bool = False,
+    ):
+        config = config or GraniteGuardianToolConfig()
+        super().__init__(config, debug)
+        self.model = config.model
+        self.default_risk_type = config.default_risk_type
+        if self.default_risk_type not in self.allowed_risk_definitions:
+            raise ValueError(
+                f"Risk check type '{self.default_risk_type}' is not supported. Allowed values are: {self.allowed_risk_definitions}",
+            )
```
NISH1001: Same:
Using pydantic field (see previous review), I don't think we'll need this check as pydantic will automatically validate the enum

NISH1001 (changes requested):

NISH1001:
@TigranTigranTigran In the PR description, could you also add one small "#Usage" section with an example. That helps with the documentation later as well. Thanks.

akd/tools/granite_guardian_tool.py
TigranTigranTigran: Done

akd/tools/granite_guardian_tool.py
TigranTigranTigran: Done

TigranTigranTigran: I can't change the term to `NSFW` because `sexual_content` is a specific definition that Granite Guardian recognizes

akd/tools/granite_guardian_tool.py
TigranTigranTigran: Done

TigranTigranTigran: Done

akd/tools/granite_guardian_tool.py
TigranTigranTigran: Done

akd/tools/granite_guardian_tool.py
TigranTigranTigran: Done

TigranTigranTigran:
Hi @NISH1001 . Thanks for your feedback! I've made the changes and updated the description with a usage section.

akd/tools/granite_guardian_tool.py
NISH1001: Eh okay.

NISH1001:
@TigranTigranTigran Thanks for the update.

One last thing (just to confirm the linting stuff):
- Could you make sure you have `pre-commit-hooks` and apply to the tool ?

```
uv pip install pre-commit-hooks
``` 

Then `pre-commit install` which should install the hook from `https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/.pre-commit-config.yaml`.

Then when you run the `git commit` it automatically should be able to run the hooks for new unstaged files.

TigranTigranTigran:
Thanks @NISH1001 

I had already installed the pre-commit, and it made changes to the commits I made

NISH1001:
I will test this locally and then we can merge.

NISH1001 (approved):

akd/tools/granite_guardian_tool.py
```diff
@@ -0,0 +1,180 @@
...
+            messages = [
+                {"role": "user", "content": item.query},
+                {"role": "assistant", "content": item.content},
+            ]
+            res = self._call_guardian(messages)
+            outputs.append(
+                {
+                    "index": idx,
+                    "query": item.query,
+                    "snippet": item.content[:100],
```
NISH1001: @TigranTigranTigran We need to change the behaviour of snippet. The hardcoded index should be avoided and we should probably utilize w.r.t to model maximum context size.

NISH1001: With the slice, it's barely getting any relevant content. Also we should probably utilize `title` as well if present.

NISH1001: Eg: 

```python
{
  "index": 1,
  "query": "landslide in nepal last year",
  "snippet": "The most adopted definition of landslide hazard combines spatial information about landslide locatio",
  "risk_label": "no",
  "is_risky": False,
  "raw_response": {
    "model": "granite3-guardian:2b",
    "created_at": "2025-07-08T14:14:25.286932Z",
    "done": True,
    "done_reason": "stop",
    "total_duration": 335484917,
    "load_duration": 4881125,
    "prompt_eval_count": 493,
    "prompt_eval_duration": 313791958,
    "eval_count": 2,
    "eval_duration": 15523417,
    "message": {
      "role": "assistant",
      "content": "No",
      "thinking": None,
      "images": None,
      "tool_calls": None
    }
  }
}
```

Here, I'd rather strategically truncate then simple character slicing.

---

The truncation here should be avoided. Let's think about this.

NISH1001: I was thinking, maybe we could just use `tiktoken` to estimate total tokens from the content (+title). And for time being, have `max_tokens` in the config to truncate the content.

```python
import tiktoken

class GraniteGuardianToolConfig(BaseToolConfig):
    """
    Configuration for Granite Guardian Tool.
    """

    model: str = "granite3-guardian"
    default_risk_type: RiskDefinition = Field(
        default=RiskDefinition.ANSWER_RELEVANCE,
        description="Default risk check",
    )
    max_tokens: int = Field(4096, description=....)
    context_buffer: int = Field(256, ...)

    @computed_field
    def tokens_to_use(self) -> int:
      return self.max_tokens - self.context_buffer


class GraniteGuardianTool(...):
  ...
  ...

  def _process_search_results(...):
    ...
    ...
     content = f"Title: {item.title}\n\n{item.content}"
    tokenizer = tiktoken.get_encoding("cl100k_base")
    tokens = tokenizer.encode(content)
    max_tokens = self.config.tokens_to_use
    if len(tokens) > self.config.max_tokens:
                  truncated_tokens = tokens[:self.config.max_tokens]
                  content = self.tokenizer.decode(truncated_tokens) + "..."
    ...
```

This way we should be able to control how many tokens to utilize for the text to be evaluated.

NISH1001: @TigranTigranTigran let's try to fix/improve this and let me know, I will review the code you push for this.

NISH1001: Need to strategically truncate or controlled truncation.

NISH1001 (changes requested):

akd/tools/granite_guardian_tool.py
NISH1001: See my previous code snippet/comment

TigranTigranTigran: @NISH1001 Thanks for your feedback. I've just made some changes along those lines, and have pushed the code. 

One thing to note though: "snippet" isn't actually being used for evaluation. The full "content" field is what's being assessed by Granite Guardian.

NISH1001: @TigranTigranTigran Oh. I thought you were truncating the full content to adjust for model's context size. Isn't that the case? Maybe I had misinterpreted the logic (if so, apology). What I was thinking was: the code was truncating the full text and that truncated text was being evaluated. If not then, maybe the method I suggested could potentially be applied to that content (before sending to the model)?

TigranTigranTigran: No problem! The truncation was just for the snippet, which isn't used for any LLM calls.

NISH1001: HOw is the max length handled for content then? The LLM should have some context limit. Maybe we can revert the commit or apply the truncation to content-level? Default to None max token -> which means no truncation....else we can truncate. For snippet, I think it's fine to have like `snippet_n_chars` or something as int in config. I was worried about context size in general. If it's not an issue, we could revert the changes maybe.

TigranTigranTigran: For the Granite Guardian model, I haven't been able to find the exact context window yet, but for the standard Granite models, it's typically around 128k tokens, so I don't anticipate any immediate issues. If we do run into any problems down the line, we can always revisit and adjust in a future PR. For now, I can revert the changes and add `snippet_n_chars` to the config to truncate the snippet. 

What do you think?

NISH1001: Ya. Reverting and adding that little snippet_n_chars is good. Let's do that and we can close the PR. Thanks

TigranTigranTigran: Thanks @NISH1001. I've made those changes and pushed the code.

TigranTigranTigran: @NISH1001 can we hold off on merging this until I've made some changes to the way Granite Guardian is called?

NISH1001: @TigranTigranTigran sure. Let me know when you think it's ready. Then we can re-review them and merge.

TigranTigranTigran: @NISH1001 Thanks - I've made the changes and have pushed the code.

akd/tools/granite_guardian_tool.py
```diff
@@ -27,6 +29,24 @@ class RiskDefinition(Enum):
     ANSWER_RELEVANCE = "answer_relevance"
 
 
+class GuardianModel_IDs(Enum):
```
NISH1001: Nitpicking: To follow standard python convention, avoid underscores in class definition. Eg: use `GuardianModelIDs` or `GuardianModelIdentifiers`

NISH1001 (changes requested):
One nitpick in class definition name.

akd/tools/granite_guardian_tool.py
TigranTigranTigran: @NISH1001 fixed, and new code pushed.

akd/tools/granite_guardian_tool.py
```diff
@@ -31,7 +31,7 @@ class RiskDefinition(Enum):
     ANSWER_RELEVANCE = "answer_relevance"
 
 
-class GuardianModel_IDs(Enum):
+class GuardianModelIDs(Enum):
```
NISH1001: Also btw [PEP 435](https://peps.python.org/pep-0435/) proposes enum to be singular name (not plular) in most of the cases.

NISH1001 (commented):
Another nitpick i forgot to add in my previous comment.

NISH1001:
@TigranTigranTigran In the usage section in the PR, does it need updated? Just wanted to confirm. Thanks for the PR thought.

akd/tools/granite_guardian_tool.py
TigranTigranTigran: Thanks, @NISH1001. That's fixed now, and I've pushed the code.

NISH1001:
I will try this tomorrow and test it out. Codewise everythign looks clean now. Thank you very much for addressing my comments. I will merge tomorrow.

akd/tools/granite_guardian_tool.py
NISH1001: Thanks

TigranTigranTigran:
@NISH1001 You're right - the usage section needed to be updated - thanks. That's done now.

akd/tools/granite_guardian_tool.py
NISH1001: Thanks. looks good.

NISH1001:
@TigranTigranTigran 
image

When does this "Missing query or content" happen?

TigranTigranTigran:
@NISH1001  That's there to  deal with the possibility of a `SearchResultItem`  not being a valid instance of the `SearchResultItem` class, i.e., not containing query and/or content  attributes.

I'm not sure if/when this would happen.

NISH1001 (approved):

NISH1001:
@TigranTigranTigran THis is working good locally. But a side note / todo list:
- In the future, we might need to use [litellm](https://github.com/BerriAI/litellm). Chances are we will have a consistent model calling in the akd code.

PR #62, opened 2025-07-03
Remove FactReasoner tool and subtree from the akd codebase

jbrry:
### Summary :memo:
This PR removes the FactReasoner (FR) tool and subtree from the repo.

### Details
The main idea will be that FR will be hosted on a cluster with endpoints for fact checking and correcting. This means that the main codebase can be removed from the repo.

In the future, when the FR service is ready, we can reinstate a tool which which will call the service with the appropriate payload.

We can also pass a list of Langchain `Documents` to the service in a future PR to index our own documents; this has been mentioned as future work.


NISH1001:
@jbrry THanks for this PR. This will remove the `akd.tools.fact_reasoner` completely, right?

jbrry:
Yes I applied `git rm -rf fact_reasoner` so should recursively remove anything below that path

NISH1001:
Cool. LGTM (Hopefully nothing breaks).

NISH1001 (approved):

PR #64, opened 2025-07-09
Feature/code search tool

pranath-reddy:
### Summary :memo:
This PR adds Initial implementation of two code search tools.
- A semantic code search tool for querying local repositories using embedding-based similarity.
- A GitHub code search tool that wraps the SearxNG tool to query public GitHub repositories via keyword-based search.

#### `LocalRepoCodeSearchTool` (Semantic Search over Local Repositories)
- Introduced `LocalRepoCodeSearchTool` under `akd.tools.code_search` to perform semantic search over local repositories.
- Uses `Embedder` from `akd.tools.misc` to compute the embeddings.
- Automatically downloads `.csv` data file from Google Drive if missing.
- Adheres to the AKD tool format.

#### `GitHubCodeSearchTool` (Keyword-Based Search via SearxNG)
- Configures a `SearxNGSearchTool` instance restricted to the GitHub engine and `technology` category.
- Allows querying GitHub repositories via keyword search using the endpoint [search/repositories](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-repositories).
- Included `examples/code_search_test.py` to test tools with custom queries asynchronously.

#### LocalRepoCodeSearchTool

```python
    LocalRepoCodeSearchToolConfig,
    CodeSearchToolInputSchema,
    LocalRepoCodeSearchTool
)

# Configure the tool
search_cfg = LocalRepoCodeSearchToolConfig(debug=True)

# Initialize the local semantic code search tool
search_tool = LocalRepoCodeSearchTool(config=search_cfg)

# Run the semantic search
    CodeSearchToolInputSchema(
        queries=["landslide nepal"],
        max_results=5
    )
)
```

#### GitHubCodeSearchTool

```python
    GitHubCodeSearchTool,
    CodeSearchToolInputSchema
)
    SearxNGSearchToolConfig
)

# Configure the GitHub code search tool
search_cfg = SearxNGSearchToolConfig(score_cutoff=0.1)

# Initialize the GitHub search tool
search_tool = GitHubCodeSearchTool(config=search_cfg)

# Run the GitHub code search
    CodeSearchToolInputSchema(
        queries=["landslide nepal"],
        max_results=10
    )
)
```

-

### Checks
- [ ] Closed #798
- [ ] Tested Changes
- [ ] Stakeholder Approval


akd/tools/code_search.py
```diff
@@ -0,0 +1,242 @@
...
+        description="The maximum number of repository results to return.",
+    )
+
+class CodeSearchResultItem(BaseModel):
+    """Schema for a single repository search result item."""
+    url: str = Field(..., description="The URL of the found repository.")
+    content: str = Field(
+        ...,
+        description="A snippet of content or description from the repository.",
+    )
```
NISH1001: Can we reuse the `akd.structures.SearchResultItem` here? Because technically it'd be the same? If not: move this to `akd.structures.CodeSearchResultItem` which directly inherits from `SearchResultItem`.

akd/tools/code_search.py
```diff
@@ -0,0 +1,242 @@
...
+from pydantic.networks import HttpUrl
+import os
+import gdown
+import pandas as pd
+
+from akd._base import InputSchema, OutputSchema
+from akd.tools._base import BaseTool, BaseToolConfig
+
+from akd.tools.gcd.embeddings import Embedder
+from akd.tools.gcd.vectorizer import RepoFinder
```
NISH1001: Let's move gcd to `akd.tools.misc` subpackage as the usages of these are just internally exposed

akd/tools/code_search.py
```diff
@@ -0,0 +1,242 @@
...
+from akd.tools.search import SearxNGSearchTool, SearxNGSearchToolConfig, SearxNGSearchToolInputSchema, SearxNGSearchToolOutputSchema
+
+class CodeSearchToolInputSchema(InputSchema):
```
NISH1001: Maybe 
```python

class CodeSearchToolInputSchema(SearchToolInputSchema):
...
```

would be better inheritance pattern.

akd/tools/code_search.py
```diff
@@ -0,0 +1,242 @@
...
+
+class CodeSearchResultItem(BaseModel):
+    """Schema for a single repository search result item."""
+    url: str = Field(..., description="The URL of the found repository.")
+    content: str = Field(
+        ...,
+        description="A snippet of content or description from the repository.",
+    )
+
+class CodeSearchOutputSchema(OutputSchema):
```
NISH1001: Maybe we can reuse `SearchToolOutputSchema` here?

akd/tools/code_search.py
```diff
@@ -0,0 +1,242 @@
...
+    """Output schema for the repository search tool."""
+    results: List[CodeSearchResultItem] = Field(
+        ...,
+        description="A list of repository search result items aggregated from all queries.",
+    )
+
+class CodeSearchToolConfig(BaseToolConfig):
+    """Configuration for the repository search tool."""
+    data_file: str = os.getenv("CODE_SEARCH_DATA_FILE", ".code_data/repositories_with_embeddings.csv") # TODO: Replace with SDE API in future.
```
NISH1001: make the `.code_data/` folder to `docs/` directly for now and reference it via akd root path.

There's `akd.utils.get_akd_root` function that reutns the root path.

For reference, you can check `akd.tools.source_validator.SourceValidatorConfig.whitelist_file_path`

akd/tools/code_search.py
```diff
@@ -0,0 +1,242 @@
...
+    config_schema = CodeSearchToolConfig
+
+    def __init__(self, config: CodeSearchToolConfig | None = None, debug: bool = False):
+        """
+        Initializes the tool. If the data file is not found, it will be
+        downloaded from Google Drive before loading the models.
+        """
+        config = config or self.config_schema()
+        super().__init__(config, debug)
+        self.repo_finder = None
```
NISH1001: can we consolidate repo finder to this class itself? the logic. And we don't have to maintain 2 different things in akd. Have most of the logic as repo finder, but no need to have repo finder as separate thing. 

And Have CodeSearchTool as a base type. and have `LocalCodeRepoCodeSearchTool` or something.


```python
class CodeSearchTool(BaseTool[CodeSearchToolInputSchema, CodeSearchOutputSchema]):
    """
    Tool for performing semantic code search. 
    It automatically downloads the necessary data file if it's not found locally.
    """
    input_schema = CodeSearchToolInputSchema
    output_schema = CodeSearchOutputSchema
    config_schema = CodeSearchToolConfig

    def __init__(self, config: CodeSearchToolConfig | None = None, debug: bool = False):
    ...

class LocalRepoCodeSearchTool(...):
  ...
```

This way, when we build more code search tool that's better, we can just have it typed to `CodeSearchTool`

NISH1001 (changes requested):

muthukumaranR:
can we also add corresponding `Node` (Agent) for this Tool? @pranath-reddy @NISH1001

akd/tools/code_search.py
pranath-reddy: Will add the base `CodeSearchTool` after I get the SDE code search tool ready. The other changes are done.

NISH1001:
@muthukumaranR let's add node as a separate pr. don't want to complicate the pr with mutlple functionalities

akd/tools/code_search.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    Input schema for the repository search tool.
+    """
+
+    top_k: int = Field(
+        10,
+        description="The maximum number of repository results to return.",
+    )
+
+class LocalRepoCodeSearchToolConfig(BaseToolConfig):
```
NISH1001: Can we do this:
Have `CodeSearchTool` as a base class and inherit any downstream codesearch tool from it? It will help keep the semantics better because we will have multiple codesearch tool. This also avoids polluting the upper hierarchy like `SearchTool`.

akd/tools/code_search.py
```diff
@@ -0,0 +1,291 @@
...
+        if not os.path.exists(data_file_path):
+            logger.warning(f"Data file not found at '{data_file_path}'. Downloading...")
+            
+            # Ensure the target directory exists
+            data_dir = os.path.dirname(data_file_path)
+            if data_dir:
+                os.makedirs(data_dir, exist_ok=True)
+            
+            # Download from Google Drive
+            file_id = "1nPaEWD9Wuf115aEmqJQusCvJlPc7AP7O"
```
NISH1001: We need to put this file information in config. We don't want to hardcode it. Have it in config.

akd/tools/code_search.py
```diff
@@ -0,0 +1,291 @@
...
+            gdown.download(f"https://drive.google.com/uc?id={file_id}", data_file_path, quiet=False)
```
NISH1001: Put the google drive download thing in `akd.utils` for separation of concerns.

akd/tools/code_search.py
```diff
@@ -0,0 +1,291 @@
...
+            for result in all_results_data
+        ]
+        
+        return self.output_schema(results=formatted_results)
+    
+"""
+Tool for performing targeted searches on GitHub using SearxNG.
+This tool is a wrapper around SearxNGSearchTool.
+"""
+class GitHubCodeSearchTool(SearxNGSearchTool):
```
NISH1001: Once we have `CodeSearchTool` as a bae, we can also do multiple inheritance for this. In python the resolution happens from right to left manner for multiple inheritance.

eg:
```python
class GitHubCodeSearchTool(CodeSearchTool, SearxNGSearchTool)
...
```
```

**Review by NISH1001 — CHANGES_REQUESTED** [review]

**Comment by NISH1001** [comment]
@pranath-reddy I think the usage section in the description here also needs to be updated? 

Thanks for the PR though. It's taking shape.

File: akd/tools/code_search.py [hunk 11]
```diff
@@ -0,0 +1,317 @@
...
+            logger.warning(f"Data file not found at '{data_file_path}'. Downloading...")
+            
+            # Ensure the target directory exists
+            data_dir = os.path.dirname(data_file_path)
+            if data_dir:
+                os.makedirs(data_dir, exist_ok=True)
+            
+            # Download from Google Drive
+            file_id = self.config.google_drive_file_id
+            google_drive_downloader(file_id, data_file_path, quiet=False)
```
NISH1001, line 178: can you add logger.debug mode statement if debug is turned on to make sure why the initialization is taking longer when I initialize the code search tool.

NISH1001 (changes requested):

akd/tools/code_search.py
NISH1001: If it's there, it's fine. Just making a comment.

akd/tools/code_search.py
```diff
@@ -0,0 +1,317 @@
...
+    config_schema = CodeSearchToolConfig
+
+"""
+Tool for performing semantic code search for local code repositories. 
+"""
+class LocalRepoCodeSearchToolInputSchema(CodeSearchToolInputSchema):
+    """
+    Input schema for the local repository code search tool.
+    """
+    top_k: int = Field(
```
NISH1001: can we make this top_k as `computed_field` and return `max_results`?

```python
from pydantic import computed_field

....
@computed_field
def top_k(self) -> int:
    return self.max_results
```

NISH1001 (changes requested):

NISH1001:
@pranath-reddy I have pushed one change with following IO validation
- The CodeSearchToolInput schema is also allowed in any codesearch tool since they have same semantics
- Have overridden `_validate_input` and `_validate_output`

akd/tools/code_search.py
```diff
@@ -0,0 +1,395 @@
...
+    """
+
+    input_schema = CodeSearchToolInputSchema
+    output_schema = CodeSearchToolOutputSchema
+    config_schema = CodeSearchToolConfig
+
+    def _validate_input(
+        self,
+        params: CodeSearchToolInputSchema | SearchToolInputSchema | dict,
+    ) -> CodeSearchToolInputSchema:
```
NISH1001, line 60: This allows us to have any input semantics as long as they are codesearch input schema or just search input schema

akd/tools/code_search.py
```diff
@@ -0,0 +1,395 @@
...
+        else:
+            raise TypeError(
+                f"params must be an instance of {self.input_schema.__name__}",
+            )
+        return params
+
+    def _validate_output(
+        self,
+        output: CodeSearchToolOutputSchema | SearchToolOutputSchema,
+    ) -> CodeSearchToolOutputSchema:
```
NISH1001, line 86: Here we validate the output and type check if it also is searrchtooloutput schema because in the `_arun` we're calling `super()` which gives output of type `SearxNGSearchToolOutputSchema` which is semantically the same as any search tool output schema

NISH1001 (approved):

PR #69, opened 2025-07-16
Remove api key error

iamsims:
In SemanticScholarSearchToolConfig, there was mismatch in the type of api_key returned by validator, it now returns the api key as a plain string


NISH1001:
@iamsims seems like this has merge conflicts. can you resolve that first?

NISH1001 (approved):

PR #71, opened 2025-07-16
Bugfix local repository search and sort the results by score

NISH1001:
### Summary :memo:
_Write an overview about it._

This pull request introduces two main improvements to the local repository code search functionality. First, it resolves a bug in the embedding parsing logic that occurred when handling non-string or NaN-like values. Second, it adds a feature to sort the search results by their relevance score, ensuring the most relevant code snippets appear first.

### Details
1.  **`akd/tools/misc.py`**:
    * Refactored the `_parse_embedding` method to remove the dependency on `pandas`.
    * Replaced the `pd.isna()` check with more robust handling for `None`, empty strings, and numpy arrays to prevent parsing errors.
2.  **`akd/tools/code_search.py`**:
    * Added a new `_sort_results` method to sort search results by score in descending order.
    * Applied this sorting to the `LocalRepoCodeSearchTool` results.
    * Improved result formatting with a more descriptive title (`f"GitHub Repository for {query}"`) and safer dictionary key access using `.pop()`.
    * Added `category="technology"` to the output schema.

### Bugfixes :bug:
- Fixed an issue in `_parse_embedding` where a `TypeError` could occur when checking for NaN values on non-DataFrame inputs. The logic now correctly handles various empty or invalid inputs without relying on pandas.

### Checks
- [x] Tested Changes

### DIFF
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -243,6 +243,41 @@ class LocalRepoCodeSearchTool(CodeSearchTool):
 
         return results.reset_index(drop=True).to_dict("records")
 
+    def _sort_results(
+        self,
+        results: list[SearchResultItem],
+        sort_by: str = "score",
+    ) -> list[SearchResultItem]:
+        """
+        Sort results by the specified key. First checks for the key directly in the dict,
+        then checks in the 'extra' field if it exists. Returns unsorted if key not found.
+        """
+
+        def __get_sort_key(result):
+            # First check if sort_by key exists directly in the dict
+            if sort_by in result:
+                return result[sort_by]
+
+            # Then check if 'extra' field exists and contains the sort_by key
+            if (
+                "extra" in result
+                and isinstance(result["extra"], dict)
+                and sort_by in result["extra"]
+            ):
+                return result["extra"][sort_by]
+
+            # If key not found anywhere, return a default value that will sort last
+            # Using float('inf') for numerical sorting or empty string for string sorting
+            return float("inf")
+
+        try:
+            # Sort in descending order (highest score first)
+            # Change reverse=False if you want ascending order
+            return sorted(results, key=__get_sort_key, reverse=True)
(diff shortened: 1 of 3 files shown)


akd/tools/code_search.py
```diff
@@ -99,6 +99,41 @@ def _validate_output(
...
+
+            # If key not found anywhere, return a default value that will sort last
+            # Using float('inf') for numerical sorting or empty string for string sorting
+            return float("inf")
+
+        try:
+            # Sort in descending order (highest score first)
+            # Change reverse=False if you want ascending order
+            return sorted(results, key=__get_sort_key, reverse=True)
+        except TypeError:
```
muthukumaranR: The the order of sort is descending but the default value if key is not found is "inf". will need to change it to -inf to not have results without scores pop up first.

akd/tools/code_search.py
```diff
@@ -272,15 +307,23 @@ async def _arun(
...
+        try:
+            formatted_results = self._sort_results(
+                formatted_results,
+                sort_by="score",
+            )
+        except Exception as e:
+            logger.error(f"Error sorting repo list by score: {e}")
 
-        return self.output_schema(results=formatted_results)
+        return self.output_schema(results=formatted_results, category="technology")
```
muthukumaranR, line 333: It's useful for mapper, but since it's not merged with develop yet.  is there another reason you have this here ?`category="technology"`? just curious how else is this being used.

muthukumaranR (changes requested):

akd/tools/code_search.py
NISH1001: this is for data structure. previously it was just adding "general". it's an artficat of the data structure returned by the search tools (like searxng)

NISH1001 (commented):
Have addressed teh changes @muthukumaranR

muthukumaranR (approved):

PR #72, opened 2025-07-17
Refactor local repository search

pranath-reddy:
### Summary :memo:
This pull request refactors the `LocalRepoCodeSearchTool` to optimize the initialization process and streamline embedding handling. It moves embedding-related preprocessing steps (embedding generation, parsing, and matrix stacking) to the constructor so they are computed once instead of during each search. 

### Details
**`akd/tools/code_search.py`**
   - **Updated the `__init__` method** of `LocalRepoCodeSearchTool`:
     - Ensures the data file is downloaded and loaded from disk.
     - Checks whether the embeddings column exists in the data.
     - If missing, invokes `generate_embeddings()` to compute and save embeddings.
     - Parses string-formatted embeddings into NumPy arrays.
     - Constructs a single stacked matrix (`self.embeddings_matrix`) from all embeddings for efficient vectorized search.
   - **Introduced the `generate_embeddings()` method**:
     - Called only if the `embeddings_column` is missing from the repo data.
     - Generates batched embeddings using the configured model.
     - Converts arrays to comma-separated strings for CSV persistence.
   - **Simplified `find_repo()`**:
     - Removed redundant parsing and matrix construction.
     - Now uses the precomputed `self.embeddings_matrix` and cached `self.repo_data`.

-

### Checks
- [ ] Closed #798
- [ ] Tested Changes
- [ ] Stakeholder Approval

### DIFF
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -158,6 +158,8 @@ class LocalRepoCodeSearchToolConfig(CodeSearchToolConfig):
     )
     embedding_model_name: str = os.getenv("CODE_SEARCH_MODEL", "all-MiniLM-L6-v2")
     remove_embedding_column: bool = True
+    text_column: str = "text"
+    embeddings_column: str = "embeddings"
     debug: bool = False
 
 
@@ -180,6 +182,7 @@ class LocalRepoCodeSearchTool(CodeSearchTool):
         Initializes the tool. If the data file is not found, it will be
         downloaded from Google Drive before loading the models.
         """
+
         config = config or self.config_schema()
         super().__init__(config, debug)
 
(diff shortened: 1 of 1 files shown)


NISH1001 (approved):
It doesn't change the semantics, correct? Overall looks good.

NISH1001:
@pranath-reddy Also just a FYI:
In case we don't want to overload the constructor `__init__`, we also can override `_post_init(...)` directly in sub classes that gets triggered after the constructor. Will help in keeping the constructor clean (if the parameters don't change)

pranath-reddy:
> It doesn't change the semantics, correct? Overall looks good.

Yes, semantics remain unchanged

PR #73, opened 2025-07-17
Add SDE Code Search Tool for API-based code search 

pranath-reddy:
### Summary :memo:
This PR adds the initial implementation of the `SDECodeSearchTool`, which queries the SDE API for relevant code repositories. It also includes a suite of PyTests for tool validation.

#### `SDECodeSearchTool` (API-Based Search via NASA SDE)
- Introduced `SDECodeSearchTool` under `akd.tools.code_search`, a tool for querying repositories via the NASA SDE code search REST API.
- Implements pagination, configurable via `max_pages` and `page_size`.
- Accepts search queries and returns parsed results in the standard `SearchResultItem` format.
- Uses `CodeSearchToolInputSchema` and `CodeSearchToolOutputSchema` to ensure compatibility with the base tool interface.

#### PyTests Added
- **Structure Tests**: Validate output and input schema integrity.
- **Google Drive Link Test**: Ensure data file for local tools is accessible.
- **Vectorizer Tests**: Validate the embedding model.

### Usage

SDE Code Search Tool

```python
    CodeSearchToolInputSchema,
    SDECodeSearchTool, 
    SDECodeSearchToolConfig
)

# Configure the tool
search_cfg = SDECodeSearchToolConfig(
    max_pages=2,
    page_size=10
)

# Initialize the SDE code search tool
search_tool = SDECodeSearchTool(config=search_cfg)

# Run the SDE search
    CodeSearchToolInputSchema(
        queries=["Weather prediction"],
        max_results=5
    )
)
```

Run tests with:

```bash
pytest tests/code_search_tool_test.py
```

-

### Checks
- [ ] Closed #798
- [ ] Tested Changes
- [ ] Stakeholder Approval

### DIFF
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -610,8 +610,8 @@ class SDECodeSearchTool(CodeSearchTool):
         formatted_results = [
             SearchResultItem(
                 title=f"SDE Code Search for {query}",
-                url=HttpUrlAdapter.validate_python(result["url"]),
-                content=result["full_text"],
+                url=HttpUrlAdapter.validate_python(result.pop("url", "")),
+                content=result.pop("full_text", ""),
                 query=query,
                 extra=result,
             )
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -406,7 +406,7 @@ class SDECodeSearchToolConfig(CodeSearchToolConfig):
     base_url: str = os.getenv("SDE_BASE_URL", "https://d2kqty7z3q8ugg.cloudfront.net/api/code/search")  
     page_size: int = 10
     max_pages: int = 1
-    debug: bool = True
+    debug: bool = False
 
 
 class SDECodeSearchTool(CodeSearchTool):
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -232,7 +232,6 @@ class LocalRepoCodeSearchTool(CodeSearchTool):
     def _ensure_data_file_exists(self):
         """
         Checks if the data file exists and downloads it if it does not.
-        Will be replaced with SDE API in future.
         """
 
         data_file_path = self.config.data_file
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -8,9 +8,10 @@ import json
 import numpy as np
 import pandas as pd
 from loguru import logger
-from pydantic import ValidationError, computed_field
+from pydantic import ValidationError, computed_field, Field
 from pydantic.networks import HttpUrl
 from scipy.spatial.distance import cdist
+from tenacity import retry, stop_after_attempt
 
 from akd.errors import SchemaValidationError
 from akd.structures import SearchResultItem
@@ -208,15 +209,21 @@ class LocalRepoCodeSearchTool(CodeSearchTool):
 
             # Parse embeddings if they are in string format
             if self.debug:
-                logger.debug(f"Embeddings column dtype: {self.repo_data[self.config.embeddings_column].dtype}")
+                logger.debug(
+                    f"Embeddings column dtype: {self.repo_data[self.config.embeddings_column].dtype}"
+                )
             if self.repo_data[self.config.embeddings_column].dtype == "object":
-                self.repo_data[self.config.embeddings_column] = self.repo_data[self.config.embeddings_column].apply(
+                self.repo_data[self.config.embeddings_column] = self.repo_data[
+                    self.config.embeddings_column
+                ].apply(
                     self.embedder._parse_embedding,
                 )
 
             # Stack all embeddings into a matrix
             if self.debug:
-                logger.debug(f"Embeddings column: {self.repo_data[self.config.embeddings_column].head()}")
+                logger.debug(
+                    f"Embeddings column: {self.repo_data[self.config.embeddings_column].head()}"
+                )
             self.embeddings_matrix = np.vstack(
                 self.repo_data[self.config.embeddings_column].tolist(),
             )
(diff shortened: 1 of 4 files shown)


akd/tools/code_search.py
```diff
@@ -443,3 +510,97 @@ async def _arun(
...
+        """
+
+        self.headers = {
+            "Content-Type": "application/json",
+            "Accept": "application/json"
+        }
+        self.base_url = config.base_url 
+        self.page_size = config.page_size
+        self.max_pages = config.max_pages
+        self.debug = config.debug
```
NISH1001: btw, this value is automatically set. no need to rewrite. This is already done at base class level.

Basically we don't need to define `__init__` as it already is done by the base class. Remove if possible. If you check [_set_attrs_from_config](https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/akd/_base.py#L166), it's already set.

akd/tools/code_search.py
```diff
@@ -443,3 +510,97 @@ async def _arun(
...
+        debug: bool = False,
+    ):
+        """
+        Initializes the SDECodeSearchTool.
+        """
+
+        self.headers = {
+            "Content-Type": "application/json",
+            "Accept": "application/json"
+        }
```
NISH1001: Have headers as part of config as well and default to this value. Will help in customization later.

akd/tools/code_search.py
```diff
@@ -443,3 +510,97 @@ async def _arun(
...
+        self.base_url = config.base_url 
+        self.page_size = config.page_size
+        self.max_pages = config.max_pages
+        self.debug = config.debug
+        super().__init__(config, debug)
+
+    def sde_search(self, page: int, query: str):
```
NISH1001, line 557: Need to probably use retrying mechanism using tenacity.
https://github.com/jd/tenacity

We can default to 2 retries.

akd/tools/code_search.py
```diff
@@ -443,3 +510,97 @@ async def _arun(
...
+        """
+
+        all_results_data = []
+        for query in params.queries:
+            if self.debug:
+                logger.debug(f"Searching for query: '{query}' with top_k={params.max_results}")
+
+            try:
+                for page in range(self.max_pages):
+                    results = self.sde_search(page=page, query=query)
```
NISH1001: Shoud probably catch exceptoin here as you're doing `response.json()["Documents"]`. IF server response fails, it will throw error. If we add exception per page level, we can skip that page call as well and track the failure and show in debug log for now.

akd/tools/code_search.py
```diff
@@ -443,3 +510,97 @@ async def _arun(
...
+
+        formatted_results = [
+            SearchResultItem(
+                title="SDE Code Search",
+                url=HttpUrlAdapter.validate_python(result["readme_url"]),
+                content=result["full_text"],
+                query=query,
+            )
+            for result in all_results_data
+        ]
```
NISH1001: Have this similar to local code search? Also sort.

examples/code_search_test.py
```diff
@@ -49,12 +61,37 @@ async def github_search_test():
         print(result.url)
         print(result.content)
         print("-"*100)
+ 
+
+async def sde_search_test():
```
NISH1001, line 64: This probably needs to breakdown. Probalby class-based test is better. Some tests for SDE is:
- check for pinging the server. If it's up or down. (If down, will probably throw some unreachable error)
- Check if the return document is confirming to the output schema

tests/code_search_tool_test.py
```diff
@@ -0,0 +1,121 @@
...
+        queries=["flood detection"],
+        max_results=10,
+    )
+    # Input structure validation
+    assert input_params.queries == ["flood detection"]
+    assert input_params.max_results == 10
+
+    # Output structure validation
+    output = await github_tool._arun(input_params)
+    validate_output_structure(output)
```
NISH1001, line 141: Let's refactor the test as metnioned above? In the github code search, probably test for searxng server is also needed. User will know if searxng is setup or not.

tests/code_search_tool_test.py
```diff
@@ -0,0 +1,121 @@
...
+        queries=["weather prediction"],
+        max_results=5,
+    )
+    # Input structure validation
+    assert input_params.queries == ["weather prediction"]
+    assert input_params.max_results == 5
+
+    # Output structure validation
+    output = await sde_tool._arun(input_params)
+    validate_output_structure(output)
```
NISH1001, line 174: Here like above: test for server as well.

tests/code_search_tool_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+    for result in output.results:
+        assert hasattr(result, "url")
+        assert hasattr(result, "content")
+        assert result.content and result.content.strip()
+
+
+# Initialize the tools and run the tests
+@pytest.fixture
+def local_tool():
```
NISH1001, line 44: For this, can we also have test for loading the data and embeddings?

NISH1001 (changes requested):

akd/tools/code_search.py
```diff
@@ -581,26 +586,42 @@ async def _arun(
...
+                all_results_data = all_results_data[: params.max_results]
             except Exception as e:
                 logger.error(f"Error during search for query '{query}': {e}")
 
         formatted_results = [
             SearchResultItem(
-                title="SDE Code Search",
-                url=HttpUrlAdapter.validate_python(result["readme_url"]),
+                title=f"SDE Code Search for {query}",
+                url=HttpUrlAdapter.validate_python(result["url"]),
```
NISH1001: result.pop("url", None)

akd/tools/code_search.py
```diff
@@ -581,26 +586,42 @@ async def _arun(
...
                 content=result["full_text"],
```
NISH1001: result.pop... because we're adding the remainder to extra...so that same information will not repeated. Hence, pop

NISH1001 (changes requested):

NISH1001:
@pranath-reddy this branch seems to have merge conflicts. can you check?

examples/code_search_test.py
pranath-reddy: `examples/code_search_test.py` are usage examples (should probably rename it). Will add the tests in `tests/code_search_tool_test.py`

NISH1001: Oh okay. I later realized it. Thanks for the updates.

NISH1001 (approved):

PR #76, opened 2025-07-21
Additional Semantic Scholar API

movinam:
### Summary:

This PR adds an additional endpoint to the SemanticScholarSearchTool. It also adds support for additional Semantic Scholar APIs in the future.

### Details:

1. Add a field named endpoint to allow arun to direct Semantic Scholar tool calls to:
a. search by default (previous functionality)
b. get paper details using DOI via the /graph/v1/paper/DOI endpoint

2. Add PaperDataItem to [structures.py](http://structures.py/). It represents a schema replicating the response from semantic scholar for a single paper ([ref](https://api.semanticscholar.org/api-docs/graph#tag/Paper-Data/operation/get_graph_get_paper)).

3. Add SearchItem as an abstraction to support different types of search results instead of restricting results to SearchResultItem. This way we can inherit from SearchItem and create result schemas based on the search. This change has been made only for SemanticScholarSearchTool as it has different endpoints that produce different results. SearchResultItem and PaperDataItem will inherit from SearchItem.

### Usage:
Will allow users to fetch paper details for a given DOI or list of DOIs by calling Semantic Scholar's DOI endpoint. 

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval


movinam:
@NISH1001  Please test and make the necessary changes required to make the PR adhere the current abstraction.

akd/tools/search.py
```diff
@@ -560,6 +564,130 @@ async def _fetch_search_page(
...
+                publication_types=item.get("publicationTypes") or None,
+                publication_date=item.get("publicationDate") or None,
+                journal=item.get("journal") or None,
+                citation_styles=item.get("citationStyles") or None,
+                authors=item.get("authors") or None,
+                citations=item.get("citations") or None,
+                references=item.get("references") or None,
+                embedding=item.get("embedding") or None,
+                tldr=item.get("tldr") or None,
+                doi=doi or None,
```
NISH1001: No need to put `or None` as `.get(<key>)` will handle None automatically.

akd/tools/search.py
```diff
@@ -791,36 +919,53 @@ async def _arun(
...
+                        query,
+                        params.category,
+                        target_results_per_query,
+                    )
+                    for query in params.queries
+                ]
+                results_per_query = await asyncio.gather(*tasks)
 
-        all_raw_results = [item for sublist in results_per_query for item in sublist]
+            all_raw_results = [item for sublist in results_per_query for item in sublist]
```
NISH1001: Can we break this into 2 different internal methods? That way this method will remain clean and easier to understand.

It's also something like a state-machine design pattern.

NISH1001 (changes requested):

akd/tools/search.py
NISH1001: Also, it's overpopulating the method. Break down into different methods and call accordingly.

akd/tools/search.py
```diff
@@ -45,7 +45,7 @@ class SearchToolOutputSchema(OutputSchema):
     """Schema for output of a tool for searching for information,
     news, references, and other content."""
 
-    results: List[SearchResultItem] = Field(
+    results: List[SearchItem] = Field(
```
NISH1001: Can we not do this. SearchResultItem.

And have doi-to-paper as a separate method in the tool `doi_to_paper` that returns the `PaperDataItem` itself.

The main tool will restul list[searchresultitem].

akd/tools/search.py
```diff
@@ -401,6 +401,8 @@ class SemanticScholarSearchToolConfig(BaseToolConfig):
         default=(os.getenv("SEMANTIC_SCHOLAR_API_KEY") or os.getenv("S2_API_KEY")),
     )
     base_url: HttpUrl = Field(default="https://api.semanticscholar.org")
+    endpoint: str = Field(default="search",
```
NISH1001: Can we separate the 2 endpoints. I want the tool to follow the same search semantics as any search tool. Default arun should and will always be search.

For doi, have a `doi_to_paper(...)` method.

Don't change the semantics of the tool for this specific purpose. If the doi to paper is needed, as as a separate method in this tool which other components can mindfully use.

akd/tools/search.py
```diff
@@ -791,36 +919,53 @@ async def _arun(
...
+                        session,
+                        query,
+                    )
+                    for query in params.queries
+                ]
+                results_per_query = await asyncio.gather(*tasks)
+            all_raw_results = [item for sublist in results_per_query for item in sublist]
+            return SemanticScholarSearchToolOutputSchema(
+                results=all_raw_results,
+            )
```
NISH1001: Put this as a separate method (doi_to_paper) without changing the original `_arun` semantics please!

akd/tools/search.py
NISH1001: The implication is a big if we change the return type. Because other search tools expect SearchResultItem which makes this tool incompaitble (loosely). I dont' want to modify existing IO semantics just for this feature.

Again, have this as a method which your knowledge gap can use directly calling:

```python

tool = SemanticScholarSearchTool(...)

paper_details = await tool.dois_to_paper_details(<list of doi>>) # and this clearly makes it apparent that the method will be getting paper details

search_items = await tool.arun(...) # see here: the search semantics still is the same: This avoids having an if-else condition within the arun method and avoids overpopulation the responsibility of this method. Single Responsiblity principle. In future if we add more features, we will delegate them to separate method
``

NISH1001 (changes requested):

NISH1001:
ONce these changes are addressed, i can have a final look and merge. Thanks for the PR.

NISH1001:
@movinam also please add test cases in `tests/` as well.

And in this PR, add the `usage` section as well. This will give idea what the new feature is doing.

Example PR:
https://github.com/NASA-IMPACT/accelerated-discovery/pull/77

akd/tools/search.py
movinam: The default will still be search, when the user specifies the endpoint as doi, it calls the doi API.

NISH1001: I know, let's not modify. Let's have a separate method. Because there are side effects of chanign types for data structure everywhere. It's not a standalone tool in the overall akd workflow. For instance, I can't have s2 tool to have different search semantics of type than other. The whole idea is that `SearchTool` will have some type semantics. And other components in the system are strictly using that type. For instance we're checking the type in some guardrails checks and other components expecting that a "search tool" will give result in a specific format and type. I also don't like type being changed based on if/else conditions. For instance, `results` will lead to list of searchresultitem or `paperdataitem` based on that if/else, which is not a great design. A function should only return a specific set of types. if there's if/else attached, then that means we need to move the added logic to a new method. This is also a single-responsibility design pattern.

WE can modify other components for the sake of adding a new feature to a tool.

This feature can itself be self-contained within `doi_to_paper_details` method.

Have `endpoint` and `doi_endpoint` as 2 different paramters.

NISH1001 (changes requested):

NISH1001 (approved):

PR #77, opened 2025-07-22
Add CombinedCodeSearchTool for unified code search 

pranath-reddy:
### Summary :memo:
This PR adds the implementation of the `CombinedCodeSearchTool`, which aggregates results from multiple code search tools and reranks them using a pretrained `CrossEncoder` model. The combined tool queries the local semantic code search, GitHub keyword-based search (via SearxNG), and the SDE API-based search, returning a unified and ranked list of relevant code repositories.

### Details

- Introduced `CombinedCodeSearchTool` under `akd.tools.code_search` to support unified querying across multiple search tools.
- Combines results from:
  - `LocalRepoCodeSearchTool` for vector similarity search over local repositories
  - `GitHubCodeSearchTool` for keyword-based GitHub repository search via SearxNG
  - `SDECodeSearchTool` for querying the NASA SDE code search API
- Aggregates outputs and reranks them using a pretrained `CrossEncoder` model (e.g., `cross-encoder/ms-marco-MiniLM-L6-v2`)
- All results are returned in the standard `SearchResultItem` format with additional metadata:
  - `extra["tool"]` stores the source tool name
  - `extra["score"]` stores the reranker score assigned by the CrossEncoder
  
### Usage

```python
    CodeSearchToolInputSchema,
    CombinedCodeSearchTool,
    CombinedCodeSearchToolConfig,
    SDECodeSearchTool,
    GitHubCodeSearchTool    
)

# Configure the tool
search_cfg = CombinedCodeSearchToolConfig()

# Initialize the code search tool
search_tool = CombinedCodeSearchTool(config=search_cfg, tools=[SDECodeSearchTool(), GitHubCodeSearchTool()])

# Run the search
    CodeSearchToolInputSchema(
        queries=["landslide nepal"],
        max_results=10
    )
)
```

### Bugfixes :bug: 
- Fixed a bug in `_sort_results` where `result["extra"]` was incorrectly used to access the `extra` field of `SearchResultItem`. It is now correctly accessed as `result.extra`.

### DIFF
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -177,14 +177,13 @@ class CombinedCodeSearchTool(CodeSearchTool):
         tools: Optional[list[CodeSearchTool]] = None,
         debug: bool = False,
     ):
-        config = config or self.config_schema()
+        super().__init__(config, debug)
         self.tools = tools or [
             LocalRepoCodeSearchTool(debug=debug),
             GitHubCodeSearchTool(debug=debug),
             SDECodeSearchTool(debug=debug),
         ]
         self.reranker_model = CrossEncoder(config.reranker_model_name)
-        super().__init__(config, debug)
 
     async def _arun(
         self,
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -32,7 +32,10 @@ class CodeSearchToolInputSchema(SearchToolInputSchema):
     Input schema for the code search tool.
     """
 
-    pass
+    @computed_field
+    def top_k(self) -> int:
+        """Returns the number of top results to return."""
+        return self.max_results
 
 
 class CodeSearchToolOutputSchema(SearchToolOutputSchema):
@@ -140,17 +143,6 @@ class CodeSearchTool(BaseTool[CodeSearchToolInputSchema, CodeSearchToolOutputSch
             return results
 
 
-class CombinedCodeSearchToolInputSchema(CodeSearchToolInputSchema):
-    """
-    Input schema for the combined code search tool.
-    """
-
-    @computed_field
-    def top_k(self) -> int:
-        """Returns the number of top results to return."""
-        return self.max_results
-
-
 class CombinedCodeSearchToolConfig(CodeSearchToolConfig):
     """
     Configuration for the combined code search tool.
@@ -167,7 +159,7 @@ class CombinedCodeSearchTool(CodeSearchTool):
     Tool for performing combined code search using multiple sub-tools and CrossEncoder reranking.
     """
 
-    input_schema = CombinedCodeSearchToolInputSchema
+    input_schema = CodeSearchToolInputSchema
     output_schema = CodeSearchToolOutputSchema
     config_schema = CombinedCodeSearchToolConfig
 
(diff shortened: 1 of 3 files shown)


akd/tools/code_search.py
```diff
@@ -139,6 +140,99 @@ def __get_sort_key(result):
...
+                if self.debug:
+                    logger.debug(f"Running tool: {tool.__class__.__name__}")
+                result = await tool._arun(params)
+                for res in result.results:
+                    res.extra["tool"] = tool.__class__.__name__
+                all_results.extend(result.results)
+            except Exception as e:
+                logger.error(f"Error running tool {tool.__class__.__name__}: {e}")
+
+        reranked = self._rerank_results(all_results, params.queries[0])[: params.top_k]
```
NISH1001, line 199: Is it `params.top_k` or config? I think we should try something like:

```python
top_k = params.top_k or self.config.top_k
```

?

akd/tools/code_search.py
```diff
@@ -139,6 +140,99 @@ def __get_sort_key(result):
...
+        debug: bool = False,
+    ):
+        config = config or self.config_schema()
+        self.tools = tools or [
+            LocalRepoCodeSearchTool(debug=debug),
+            GitHubCodeSearchTool(debug=debug),
+            SDECodeSearchTool(debug=debug),
+        ]
+        self.reranker_model = CrossEncoder(config.reranker_model_name)
+        super().__init__(config, debug)
```
NISH1001: Better and non-side effect will be following order

```python
...
super().__init(...)
self.tools = ...
self.reranker_model = ...
```

Notice: We invoke base's constructor first. This way we don't have to to `config = config or self.config_schema()` because super() will automatically handle it.

NISH1001 (changes requested):

akd/tools/code_search.py
pranath-reddy: it's `params.top_k`, we don't have a top_k field in the config schema

NISH1001: I mean max_results. THere's a computed_field as well.

pranath-reddy: yes, `params.top_k` reflects `params.max_results` right?

NISH1001: Oh ya

NISH1001:
@pranath-reddy could you confirm if it can also accept the code search input schema?

NISH1001:
And in the usage section could you also add different code search tools as parameters list to let know that it's the original purpose

pranath-reddy:
> @pranath-reddy could you confirm if it can also accept the code search input schema?

It doesn't work. Should I add `top_k` as a computed field directly to `CodeSearchToolInputSchema`?

NISH1001:
> > @pranath-reddy could you confirm if it can also accept the code search input schema?
> 
> It doesn't work. Should I add `top_k` as a computed field directly to `CodeSearchToolInputSchema`?

Yes. Let's do it

akd/tools/code_search.py
```diff
@@ -232,17 +224,6 @@ def _rerank_results(
...
 
-class LocalRepoCodeSearchToolInputSchema(CodeSearchToolInputSchema):
-    """
-    Input schema for the local repository code search tool.
-    """
-
-    @computed_field
-    def top_k(self) -> int:
-        """Returns the number of top results to return."""
-        return self.max_results
```
NISH1001, line 150: Does this mean, these are removed, right?

NISH1001 (approved):

PR #78, opened 2025-07-22
Refactor the scraper module to make it robust and extensiblity

NISH1001:
### Summary :memo:

This pull request introduces a comprehensive refactoring of the scraper module to create a more unified, extensible, and robust framework. The core of this update is the introduction of a new base class for all scrapers, which standardizes their interfaces and functionality. Additionally, new specialized scraper types like **`PDFScraper`**, **`WebScraper`**, and the versatile **`OmniScraper`** have been introduced to handle different content sources more effectively.

-----

#### Major Changes

  * **New Base Scraper Framework:** Introduced **`ScraperToolBase`**, **`ScraperToolInputSchema`**, and **`ScraperToolOutputSchema`** in `akd/tools/scrapers/_base.py` to provide a common structure for all scraping tools. This now includes specialized base classes:
      * **`WebScraper`**: The base for all web-based scrapers.
      * **`PDFScraper`**: The base for handling PDF files, with helper methods to detect and process PDFs from both local and remote paths.
      * **`OmniScraper`**: A powerful new scraper type that can handle multiple document formats. Any new scraper capable of handling multiple formats should be added to the `scrapers.omni` module.
  * **New `DoclingScraper`:** Added a powerful **`DoclingScraper`** that leverages the `docling` library to handle various document formats (PDF, HTML, DOCX, etc.) with advanced parsing capabilities. For future development, any new scraper capable of handling multiple formats should be added to the `scrapers.omni` module.
  * **New `CompositeScraper`:** Implemented a **`CompositeScraper`** that runs multiple scrapers in a waterfall sequence, trying the next one if the previous one fails.
  * **Refactored Existing Scrapers:** Migrated `SimpleWebScraper` and `SimplePDFScraper` to inherit from the new base classes, ensuring consistent API and behavior.

#### Minor Changes

  * **Unified Type Hinting:** Standardized Pydantic type hints, such as replacing `HttpUrl` with the more flexible **`AnyUrl`** for broader compatibility with different URL schemes (including `file://`).
  * **Code Cleanup:** Restructured imports and improved code organization across the scraper module for better maintainability.

#### Notes
- If we want to add new pdf-only parser, we can inherit from `PDFScraper` type.
- If we want to improve multi-scrapers, use `OmniScraper` type
- If we want to add new web scrapers, use `WebScraper` type

-----

### Usage 🚀

The new `CompositeScraper` allows you to chain multiple scrapers together. It will attempt to use them in the order they are provided until one successfully returns content.

Here's how you can use it to scrape a URL, which could be a web page or a local/remote PDF:

```python
from akd.tools.scrapers import (
    ScraperToolInputSchema,
    SimpleWebScraper,
    Crawl4AIWebScraper,
    SimplePDFScraper,
    DoclingScraper,
)
from akd.tools.scrapers.composite import CompositeScraper

# The URL can point to a webpage, a remote PDF, or a local file
_url = "https://arxiv.org/pdf/2411.08181"

# Initialize the composite scraper with a sequence of scrapers to try
scraper = CompositeScraper(
    DoclingScraper(),
    Crawl4AIWebScraper(),
    SimpleWebScraper(),
    SimplePDFScraper(),
)

# Run the scraper
output = await scraper.arun(ScraperToolInputSchema(url=_url))

# View the output
print(output.content)
```

Here, the composite scraper behaves the same way as its individual counterparts semantically. So, from outside usage perspective, the IO semantics remain the same.

-----

### Checks
  - [x] Tested Changes
  - [ ] Stakeholder Approval
  - [ ] Unit Tests


akd/structures.py
```diff
@@ -31,40 +31,40 @@ class SearchResultItem(BaseModel):
     """Represents a single search result item with metadata."""
 
     # Required fields
-    url: HttpUrl = Field(..., description="The URL of the search result")
+    url: AnyUrl = Field(..., description="The URL of the search result")
```
NISH1001, line 34: Just to clarify: This doesn't have any side-effects because original `HttpUrl` is a subtype of `AnyUrl`. So, it still works and doesn't break any downstreams.

NISH1001 (commented):
Some notes:

akd/tools/scrapers/composite.py
```diff
@@ -1,33 +1,75 @@
...
+    )
 
-class CompositeScraper(WebScraperToolBase):
-    """Composite web scraper that runs multiple scrapers in sequence."""
+    from akd.tools.scrapers import DoclingScraper
+
+    from akd.tools.scrapers.composite import CompositeScraper
+
+    _url = "https://arxiv.org/pdf/2402.01822v1"
+    # _url = "https://nishparadox.com/writing/vipassana-retreat.html"
```
muthukumaranR, line 44: Error: Human communication not found

akd/tools/scrapers/composite.py
```diff
@@ -40,8 +82,11 @@ async def _arun(
                         f"Running scraper={scraper.__class__.__name__} for {params}",
                     )
 
-                result = await scraper.arun(params)
+                result = await scraper.arun(scraper.input_schema(**params.model_dump()))
```
muthukumaranR, line 85: why won't previous `result = await scraper._arun(params)` work here? just curious

akd/tools/scrapers/pdf_scrapers.py
```diff
@@ -218,20 +184,21 @@ async def _download_pdf_from_url(
...
+                    response.raise_for_status()
 
-            logger.debug(f"Downloading PDF at {temp_file.name}")
-            with open(temp_file.name, "wb") as f:
-                for chunk in response.iter_content(chunk_size=8192):
-                    f.write(chunk)
+                    logger.debug(f"Downloading PDF at {temp_file.name}")
+                    with open(temp_file.name, "wb") as f:
+                        async for chunk in response.aiter_bytes(chunk_size=8192):
+                            f.write(chunk)
```
muthukumaranR: should we implement some safety checks here rather than leaving the download unbound? for bytes, similar to WebScraperConfig, and/or timeout?

akd/tools/scrapers/composite.py
```diff
@@ -1,33 +1,75 @@
...
+
+    output = await _scraper.arun(ScraperToolInputSchema(url=_url))
+    ```
+    """
 
     def __init__(
         self,
-        *scrapers: WebScraperToolBase,
+        *scrapers: ScraperToolBase,
         debug: bool = False,
```
muthukumaranR, line 60: is this parameter used? implementation uses `self.debug`, which i assume will be from config, and set by `_set_attrs_from_config`, but is this parameter supposed to override the setting? I might have missed it.

akd/tools/scrapers/_base.py
```diff
@@ -0,0 +1,367 @@
...
+            soup (BeautifulSoup): The parsed HTML content.
+
+        Returns:
+            Optional[str]: The extracted author name(s) or None if not found.
+        """
+        # Check standard meta tags
+        author_meta = soup.find("meta", attrs={"name": "author"})
+        if author_meta and author_meta.get("content"):
+            return author_meta.get("content")
+
```
muthukumaranR, line 137: `soup.select('meta[name], ...)` might be better here since we might have multiple meta tags and `find()` returns only one

muthukumaranR (approved):

akd/tools/scrapers/composite.py
NISH1001: Oh ya. I think this composite doesn't have any config. So, debug itself is not initialized. Good catch for the bug. Thanks

NISH1001: Okay. Seems like it's already being done `super().__init__(debug=debug)` using this.

akd/tools/scrapers/pdf_scrapers.py
NISH1001: Good catch. Let me see what can be done. Also chunk-size could be also config-controlled I guess.

akd/tools/scrapers/composite.py
NISH1001: Oh ya. I can explain. Let me try:

Since each scraper could have a slightly different variation of input schema (for instance, Pdfscraper input schema uses its own modified schema derived from the base scraper input schema, with adding one `field_validator` to process whether incoming url is a local path or actual HTTP.

```python
class PDFScraperInputSchema(ScraperToolInputSchema):
    """
    Input schema for the PDFScraper tool.

    Accepts both web URLs and local file paths. File paths are automatically
    converted to file:// URLs for proper validation.

    Examples:
        - Web URL: "https://example.com/file.pdf"
        - Local file: "/Users/name/document.pdf"
        - Relative file: "./documents/file.pdf"
    """

    url: AnyUrl = Field(
        ...,
        description="The URL or local file path of the PDF to scrape.",
    )

    @field_validator("url", mode="before")
    @classmethod
    def convert_file_path_to_url(cls, v):
        """
        Convert file paths to file:// URLs for proper Pydantic validation.

        Args:
            v: Input value (URL string or file path)

        Returns:
            str: Properly formatted URL
        """
        if not isinstance(v, str):
            return v

        # Check if it's already a URL (has a scheme)
        if any(
            v.lower().startswith(scheme)
            for scheme in ["http://", "https://", "file://"]
        ):
            return v

        # Convert file path to file:// URL
        if os.path.isabs(v):
            # Absolute path
            # Handle Windows paths by converting backslashes to forward slashes
            normalized_path = v.replace("\\", "/")
            if normalized_path.startswith("/"):
                return f"file://{normalized_path}"
            else:
                # Windows absolute path like C:/path
                return f"file:///{normalized_path}"
        else:
            # Relative path
            return f"file://{v}"
```

I know it feels convoluted. Having "str" might have solved it, but I wanted to retain the type of Any Url be ith httpurl or file url / path. For `AnyUrl` to accept local file path (eg : `/home/xyz/file.pdf`), it should have a schema (eg: `http://`, `file://`). Hence, original input params will fail because this scraper is expecting its own input type.

akd/tools/scrapers/omni.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+
+    input_schema = ScraperToolInputSchema
+    output_schema = ScraperToolOutputSchema
+    config_schema = ScraperToolConfig
+
+
+class DoclingScraperConfig(ScraperToolConfig):
+    pdf_mode: Literal["accurate", "fast"] = Field(default="fast")
+    export_type: Literal["markdown", "html"] = Field(default="markdown")
```
movinam: I don't see the function for Docling to process the output when export_type = html. Could that be added as well? 

export_to_html() and export_to_markdown() return different results, it would be useful to support both export types ([ref](https://docling-project.github.io/docling/usage/supported_formats/)).

akd/tools/scrapers/_base.py
NISH1001: I don't know how to test this. DO you have any specific URLs we can test the changes on? Should we leave this for now? Title extraction has been hit and miss in some of the websites.

akd/tools/scrapers/omni.py
NISH1001: Good catch. Let me check. If not, we can have a new small PR for adding that export to HTML maybe.

NISH1001: Okay. I have added this option based on the `export_type` parameter. It's working nicely. Thanks for the suggestion.

For other: I will leave it for now. We can work on exporting in other scraper later.

PR #79, opened 2025-07-23
Add wrapper tool for FactReasoner

jbrry:
### Summary :memo:
This PR adds `FactCheckTool` which is a wrapper to call the hosted FactReasoner service.

It also adds some helper tools which enable further downstream analysis: a `VectorDBTool` which is based on a Chroma vector database is added so that search artefacts can be stored for downstream analysis or so that FactReasoner can use it for context retrieval.

### Details
`VectorDBTool`: `akd/tools/vector_db_tool.py` used for interacting with a vector database. It uses ChromaDB for persistent storage and sentence-transformers for creating embeddings. The `.index()` method is used for adding documents, and `_arun` runs the retrieval based on a query.

`DeepLitSearchAgent` Integration: [The new functionality has been added as part of a DeepLitSearchAgent run](https://github.com/NASA-IMPACT/accelerated-discovery/blob/feature/vector_database_tool/scripts/demo_deep_search_with_fact_check.py). This shows how we can use the new tools to work along with the `DeepLitSearchAgent`.

### Usage

Documentation for the backend deployment is available [here](https://factreasoner-service-app.1yhbkn094k2v.us-south.codeengine.appdomain.cloud/docs#/).

Standalone example:
```
from akd.tools.fact_check import FactCheckInputSchema, FactCheckOutputSchema, FactCheckToolConfig, FactCheckTool

from pydantic import HttpUrl

fact_check_config = FactCheckToolConfig(base_url=HttpUrl("https://factreasoner-service-app.1yhbkn094k2v.us-south.codeengine.appdomain.cloud"), polling_interval_seconds=10)
fact_check_tool = FactCheckTool(config=fact_check_config)


sample_question = "What is the coriolis effect?"
sample_answer = "The Coriolis effect describes the pattern of deflection taken by objects not firmly connected to the ground as they travel long distances around the Earth."

fact_check_input = FactCheckInputSchema(
    question=sample_question,
    answer=sample_answer,
)

print(f"\n--- Running Fact-Check on sample answer ---\n{sample_answer}\n")
try:
    fact_check_result = await fact_check_tool.arun(params=fact_check_input)

    print("\n--- Fact-Check Complete ---")
    score = fact_check_result.fact_reasoner_score.get("factuality_score", 0)
    num_supported = len(fact_check_result.supported_atoms)
    num_not_supported = len(fact_check_result.not_supported_atoms)

    print(f"Factuality Score: {score:.2%}")
    print(f"Supported Atoms: {num_supported}")
    print(f"Not Supported Atoms: {num_not_supported}")
    print(f"Graph ID: {fact_check_result.graph_id}")
    
    # graph methods
    if fact_check_result.graph_id:
        graph_id = fact_check_result.graph_id

    print(f"\nFetching graph structure for ID: {graph_id}...")
    graph_json_url = f"{fact_check_tool.config.graph_json_endpoint}/{graph_id}?format=full"
    graph_response = await fact_check_tool.api_client.get(graph_json_url)

    if graph_response.status_code == 200:
        graph_data = graph_response.json()
        print(f"Successfully fetched graph JSON. Found {len(graph_data.get('nodes', []))} nodes and {len(graph_data.get('edges', []))} edges.")
        # print(graph_data) # uncomment for full object.
    else:
        print(f"Failed to fetch graph JSON. Status: {graph_response.status_code}")

except Exception as e:
    print(f"\nExample failed. Ensure the fact-checking service is running.")
    print(f"Error: {e}")
```

It returns the factuality score based on the question and response pair:

```
--- Fact-Check Complete ---
Factuality Score: 100.00%
Supported Atoms: 3
Not Supported Atoms: 0
Graph ID: 92ed5024-ad09-48f6-b296-357bdc3e38f7

Fetching graph structure for ID: 92ed5024-ad09-48f6-b296-357bdc3e38f7...
Successfully fetched graph JSON. Found 15 nodes and 36 edges.
```

I am also listing the test files for each of these tools, which cover initialization and running of the tools and can serve as documentation:
- `VectorDBTool`: [test script example](https://github.com/NASA-IMPACT/accelerated-discovery/blob/feature/vector_database_tool/tests/tools/vector_db_tool_test.py)
- `FactCheckTool`: [test script example](https://github.com/NASA-IMPACT/accelerated-discovery/blob/feature/vector_database_tool/tests/tools/test_fact_check.py).


NISH1001:
@jbrry Could you please add `usage` section in the PR as well.

Some sample PRs:
- https://github.com/NASA-IMPACT/accelerated-discovery/pull/78
- https://github.com/NASA-IMPACT/accelerated-discovery/pull/77

etc

akd/agents/litsearch.py
```diff
@@ -71,10 +76,19 @@ def __init__(
...
 
         self.n_queries = n_queries
         super().__init__(debug=debug)
 
+    async def get_response_async(self, *args, **kwargs) -> LitAgentOutputSchema:
+        """
+        This method is required by the BaseAgent but is not used by LitAgent,
+        which is an orchestrator. The main entry point is the `_arun` method.
+        """
+        raise NotImplementedError()
```
NISH1001: I had one small todo: To make in not abstract. But glad that you figured this out.

akd/tools/text_splitter.py
```diff
@@ -0,0 +1,93 @@
...
+    config_schema = TextSplitterToolConfig
+
+    def __init__(
+        self,
+        config: Optional[TextSplitterToolConfig] = None,
+        debug: bool = False,
+    ):
+        """Initializes the TextSplitterTool."""
+        config = config or self.config_schema()
+        super().__init__(config, debug)
```
NISH1001: btw, i think `config = config or self.config_schema()` is already handled by the base when super() is invoked.

akd/tools/text_splitter.py
```diff
@@ -0,0 +1,93 @@
...
+    documents: List[Document] = Field(
+        ...,
+        description="A list of Langchain Document objects to split.",
+    )
+
+
+class TextSplitterOutputSchema(OutputSchema):
+    """Output schema for the Text Splitter Tool."""
+
+    chunks: List[Document] = Field(
```
NISH1001: Can we confirm if `Document` is serializable easily?

NISH1001 (commented):
[Initial superficial ocmments]

akd/tools/text_splitter.py
jbrry: Thanks I will use the `| None` pattern I've seen in other files!

akd/tools/text_splitter.py
jbrry: Yes it has Pydantic attributes like `model_dump`:

```
from langchain_core.documents import Document

doc = Document(page_content="The sky is blue.", metadata={"source": "weather.txt", "id": 123})

print(doc)
> Document(metadata={'source': 'weather.txt', 'id': 123}, page_content='The sky is blue.')


doc_dict = doc.model_dump()
print(doc_dict)
> {'id': None, 'metadata': {'source': 'weather.txt', 'id': 123}, 'page_content': 'The sky is blue.', 'type': 'Document'}

doc_json = doc.model_dump_json()
print(doc_json)
> '{"id":null,"metadata":{"source":"weather.txt","id":123},"page_content":"The sky is blue.","type":"Document"}'
```

tests/tools/semantic_scholar_search_test.py
movinam, line 1: Could we rename the file to test_semantic_scholar_search for readability?

akd/tools/text_splitter.py
movinam, line 1: I think it makes more sense to keep text_splitter local to the tool that uses it (in this case fact check). Since text splitters can take many different forms (as shown [here](https://api.python.langchain.com/en/latest/text_splitters_api_reference.html)), they’re not really generic. If we were to create a standalone text splitter tool, we’d need to support all those variations, which would essentially mean re-implementing an existing library.

cc @NISH1001

akd/tools/vector_db_tool.py
```diff
@@ -0,0 +1,142 @@
...
+class VectorDBOutputSchema(OutputSchema):
+    """Output schema for the Vector Database tool's query results."""
+
+    results: List[Document] = Field(
+        ...,
+        description="List of retrieved Langchain Document objects.",
+    )
+
+
+class VectorDBToolConfig(BaseToolConfig):
```
movinam, line 44: We could potentially add a default text splitter here which can be used by the index function if configured? This is an alternative to adding the text_splitter tool

akd/tools/fact_check.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+
+
+class FactCheckToolConfig(BaseToolConfig):
+    """Configuration for the FactCheckTool."""
+
+    base_url: HttpUrl = Field(
+        # default="http://localhost:8011",
+        default=HttpUrl(
+            "https://factreasoner-service-app.1yhbkn094k2v.us-south.codeengine.appdomain.cloud",
```
movinam: We should move this to configs in case it changes in the future to avoid code changes. You can set the default to local host and fetch from the environment.

akd/tools/vector_db_tool.py
```diff
@@ -0,0 +1,142 @@
...
+    embedding_model_name: str = Field(
+        default="sentence-transformers/all-MiniLM-L6-v2",
+        description="The name of the Hugging Face embedding model to use.",
+    )
+    embedding_model_api_key: Optional[str] = Field(
+        default=None,
+        description="The API key for the embedding model provider, currently using HuggingFace.",
+    )
+    db_path: str = Field(
+        default="./chroma_db",
```
movinam: This should be moved to config or made an environment variable as well. The default can be set there.

tests/tools/vector_db_tool_test.py
movinam, line 1: Can this be renamed to test_vector_db as well?

akd/tools/text_splitter.py
movinam:
```python
from langchain_text_splitters import RecursiveCharacterTextSplitter
text_splitter = RecursiveCharacterTextSplitter(chunk_size=100, chunk_overlap=0)
texts = text_splitter.split_text(document)
```
^ we could just do this and index with the VectorDB (after processing depending if needed).
Otherwise, we are building abstraction over something that already exists.

movinam (commented):
Thanks for the PR! I think the text splitter tool is not needed - we can just use the langchain implementation/or any existing implementation and pass the documents to the vector db to be indexed. Please see comments.

tests/tools/semantic_scholar_search_test.py
jbrry: Can do, I was using the existing one [granite_guardian_tool_test.py](https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/tests/tools/granite_guardian_tool_test.py) as a reference, but can settle on `test_*.py` or `*_test.py`, and go with that for all tests.

akd/tools/text_splitter.py
jbrry: Yes I agree, we can just import the splitter for generic splitting use-cases.

akd/tools/vector_db_tool.py
jbrry: Yeah I think it's a nice idea but might want to keep the two de-coupled for the time being. I will go with the approach of removing the splitter and if there is a need for it in the future, it could be added here. But in general, I would say we'd have to support different splitting methods before indexing so might end up putting too much emphasis on splitting in this `VectorDBTool` class.

akd/tools/vector_db_tool.py
jbrry: I think in general we need to have a bigger discussion about how parameters should be set.
Right now, all configurations exist in a python file, and can be overridden only using python scripts, e.g. instantiating the agent/tool in python, and initialising it with its config where certain parameters are adjusted. (Caveat, I know the storm settings utilise BaseSettings which in turn come from the .env, but I think the .env isn't the right choice to handle all of our runtime settings longterm as we have so many tools/options).

This can be problematic in cases where you want to run the exact same agent setting with different search engines. To do this now you would have to modify the script file with different settings for each run. This isn't scaleable and is error-prone.

I think we need to have a main entry point/runfile that can be ran with a yaml, json, or toml file argument. This config file argument defines the full settings of the agent run, e.g. the main agent type and its tools, and its sub-agent(s) and their tools, and so on.

The runfile then unpacks this config and builds the whole system, so it can be a very high level agent like the DeepLitSearchAgent, or a very simple one that just does query relevancy checking etc. We wouldn't need a new script file for each setting, just a modified json/yaml/toml file.

cc @NISH1001 @muthukumaranR

tests/tools/semantic_scholar_search_test.py
jbrry: renamed

akd/tools/text_splitter.py
jbrry: Splitter is applied within the script now: [ref](https://github.com/NASA-IMPACT/accelerated-discovery/blob/ae2f2a12c1d07db8f1524550ced6c9f4d05857ca/scripts/demo_deep_search_with_fact_check.py#L112)

akd/tools/vector_db_tool.py
jbrry: These path/url params are coming from the `.env` for now (using other examples like):
```
    ollama_base_url: HttpUrl = Field(
        default=HttpUrl(os.getenv("OLLAMA_BASE_URL", "http://localhost:11434")),
    )
```

-> 

```
    base_url: HttpUrl = Field(
        default=HttpUrl(os.getenv("FACT_CHECK_API_URL", "http://localhost:8011")),
        description="The base URL of the remote Fact-Checking and Correction Service.",
    )
```

NISH1001:
@jbrry 
```
FAILED tests/tools/test_fact_check.py::test_fact_check_tool_with_mock - KeyError: 'job_id'
```

This has some errors. Can we fix this?

pyproject.toml
```diff
@@ -46,7 +46,11 @@ dependencies = [
     "sentence-transformers>=5.0.0",
     "ollama>=0.5.1",
     "tiktoken>=0.9.0",
+    "rapidfuzz>=3.13.0",
     "deepeval>=3.4.0",
+    "markdown>=3.8.2",
+    "chromadb>=1.0.13",
+    "pytest-mock>=3.14.1",
```
NISH1001: ca we add pytest-related this to `project.optional-dependencies` in pyprojet instead of main dependency?

akd/tools/fact_check.py
```diff
@@ -0,0 +1,151 @@
...
+        logger.info(
+            f"Sending fact-check request for question: '{params.question[:50]}...'",
+        )
+
+        try:
+            # Start the job
+            start_response = await self.api_client.post(
+                "/fact-check/start",
+                json=params.model_dump(),
+                timeout=60.0,
```
NISH1001: Can we add the timeout as a config param as well and default to 60?

akd/tools/vector_db_tool.py
```diff
@@ -0,0 +1,143 @@
...
+    embedding_model_name: str = Field(
+        default="sentence-transformers/all-MiniLM-L6-v2",
+        description="The name of the Hugging Face embedding model to use.",
+    )
+    embedding_model_api_key: Optional[str] = Field(
+        default=os.getenv("EMBEDDING_MODEL_API_KEY", None),
+        description="The API key for the embedding model provider, if required.",
+    )
+    db_path: str = Field(
+        default=os.getenv("VECTOR_DB_PATH", "./chroma_db"),
```
NISH1001: Can we have the db path using relative to `get_akd_root()`? `from akd.utils import get_akd_root`?

NISH1001 (changes requested):

akd/tools/fact_check.py
jbrry: Good idea, all requests have this timeout now, the overall job timeout is external to the Httpx client but is also set in .env now.

pyproject.toml
jbrry: good catch, done

akd/tools/vector_db_tool.py
jbrry: Good suggestion, done!

jbrry:
> @jbrry
> 
> ```
> FAILED tests/tools/test_fact_check.py::test_fact_check_tool_with_mock - KeyError: 'job_id'
> ```
> 
> This has some errors. Can we fix this?

Fixed, the test file hadn't been updated to test the two-step approach in run.
Previously long jobs would crash on Code Engine, so the two-step approach first starts the job, which is instant, then polls it until completion/failure for certain intervals.

NISH1001:
@jbrry One last request. Can we have factchecktool usage section as code block in the PR description? That way it avoids having to navigate to the scripts. Just for factchecktool. Otherwise it LGTM

akd/tools/vector_db_tool.py
NISH1001: Is this resolved? Some examples:
https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/akd/tools/source_validator.py#L93

https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/akd/tools/code_search.py#L252

We can default to some root-level temp db no for defaults? using `get_akd_root()`?

akd/tools/vector_db_tool.py
```diff
@@ -0,0 +1,149 @@
+import os
+from typing import List, Optional
+
+import chromadb
+import chromadb.utils.embedding_functions as embedding_functions
+from langchain_core.documents import Document
```
NISH1001: Is Langchain's `Document'` dependency actually needed? Can find another way? Like `akd.structures.Document` or something with similar but minimal data structure?

NISH1001 (changes requested):

akd/tools/vector_db_tool.py
jbrry: I removed the Document dependency, we just need to pass a list of ids, texts and metadata, as expected by Chroma's `.add()` [method](https://docs.trychroma.com/docs/collections/add-data)

akd/tools/fact_check.py
```diff
@@ -0,0 +1,158 @@
...
+        Starts a fact-checking job and polls for its completion.
+        """
+        logger.info(
+            f"Sending fact-check request for question: '{params.question[:50]}...'",
+        )
+
+        try:
+            # Start the job
+            start_response = await self.api_client.post(
+                "/fact-check/start",
```
NISH1001: Can we have this endpoint also from config? In case in future decide to change the endpoint

akd/tools/fact_check.py
```diff
@@ -0,0 +1,158 @@
...
+            start_response.raise_for_status()
+            job_id = start_response.json()["job_id"]
+            logger.info(f"Successfully started job with ID: {job_id}")
+
+            # Poll for the result
+            total_wait_time = 0
+            while total_wait_time < self.config.job_timeout_seconds:
+                logger.info(f"Polling status for job {job_id}...")
+                status_response = await self.api_client.get(
+                    f"/fact-check/status/{job_id}",
```
NISH1001: same as above. Probably make this configurable and default to existing endpoint.

NISH1001 (changes requested):

NISH1001 (approved):

NISH1001:
@jbrry There's merge conflict with develop. Could you pull in develop latest changes and resolve?

NISH1001 (approved):

akd/tools/vector_db_tool.py
```diff
@@ -0,0 +1,144 @@
...
+    embedding_model_api_key: Optional[str] = Field(
+        default=os.getenv("EMBEDDING_MODEL_API_KEY", None),
+        description="The API key for the embedding model provider, if required.",
+    )
+    db_path: str = Field(
+        default=os.getenv("VECTOR_DB_PATH", str(get_akd_root() / "chroma_db")),
+        description="Path to the persistent ChromaDB directory.",
+    )
+    collection_name: str = Field(
+        default="litagent_demo",
```
NISH1001: Can we rename the default name to something else? like akd_vdb or something?

NISH1001 (commented):
Nitpick

NISH1001:
Closing this PR for time being.

PR #84, opened 2025-07-24
Modify doi_to_paper to use external ID instead

movinam:
### Summary:
This paper modifies the doi_to_paper function to fetch papers using any external ID.
By default, the code will use DOI, otherwise, it can fetch papers using these IDs - ["DOI", "ARXIV", "PMID", "ACL", "MAG", "CorpusId"]

### Usage:
```python
from akd.tools.search import SemanticScholarSearchTool, SemanticScholarSearchToolInputSchema

semantic_search_tool = SemanticScholarSearchTool()
output_with_ext_id = await semantic_search_tool.fetch_paper_by_external_id(SemanticScholarSearchToolInputSchema(queries=["2502.10522"]), external_id='arxiv')
output_with_defaults = await semantic_search_tool.fetch_paper_by_external_id(SemanticScholarSearchToolInputSchema(queries=["10.1145/3701716.3717805"]))
```

### Checks
- [ ] Tested Changes
- [ ] Stakeholder Approval

### DIFF
--- a/akd/tools/search.py
+++ b/akd/tools/search.py
@@ -431,6 +431,8 @@ class SemanticScholarSearchToolConfig(BaseToolConfig):
             ),
         ),
     )
+    # External ID to fetch paper details with
+    external_id: Optional[Literal["DOI", "ARXIV", "PMID", "ACL", "MAG", "CorpusId", "PMCID", "URL"]] = Field(default="DOI")
     debug: bool = False
 
     @field_validator("api_key", mode="before")
@@ -467,6 +469,7 @@ class SemanticScholarSearchTool(
         fields: Optional[List[str]] = None,
         results_per_page: int = 100,
         max_pages_per_query: int = 5,
+        external_id: str = "DOI",
         debug: bool = False,
     ) -> SemanticScholarSearchTool:
         """Creates an instance from specific parameters."""
@@ -476,6 +479,7 @@ class SemanticScholarSearchTool(
             "max_results": max_results,
             "results_per_page": results_per_page,
             "max_pages_per_query": max_pages_per_query,
+            "external_id": external_id,
             "debug": debug,
         }
         if fields:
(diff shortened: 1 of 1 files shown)


akd/tools/search.py
```diff
@@ -879,24 +880,34 @@ async def _process_final_results(
...
             params: Input parameters including queries and category.
-
+            **kwargs:
+            - external_id (str, optional): The type of external identifier ("DOI", "ARXIV", "PMID", "ACL", "MAG", "CorpusId", "PMCID", "URL").
+              Defaults to "DOI" if not provided.
         Returns:
             List of PaperDataItem objects.
         """
+        external_id = kwargs.get("external_id", "DOI")
+        if external_id not in ["DOI", "ARXIV", "PMID", "ACL", "MAG", "CorpusId", "PMCID", "URL"]:
```
NISH1001: can we have this list as a part of s2 config and default to this list.

NISH1001 (changes requested):

akd/tools/search.py
movinam: I would not recommend that, different URLs (or queries) would have different external IDs. It's not a configurable property. We could add it as a default to SemanticScholarSearchToolInputSchema instead.

NISH1001: I disagree. I don't recommended hard coded values in code that's not a configurable. Either set the list is internal private attributes or config. Anything hard coded is a bad design.

I'm suggesting for the list values here where you're checking. Can you not make the list hard coded . You can have the list as Semantic Scholar Search Tool dot underscore s2_external_ids

NISH1001: Don't change the input schema. Since you're hard coding, it suggests it's always the case. In future  what happens if new things in the s2 API are added. We would have to resort to changing the list values. Again: anything hard coded is highly discouraged

movinam:
fetch_paper_by_external_id allows external_id to be passed as a keyword argument to support queries with different external_ids.

NISH1001 (approved):

akd/tools/search.py
```diff
@@ -684,7 +689,7 @@ async def _fetch_paper_by_doi(
                 f"Failed to fetch Semantic Scholar results for query '{query}': {e}",
             )
 
-        return []
+        return [None]
```
NISH1001, line 692: Can we return just empty list or similar empty datastrcutreu? because it will fail in some downtram logic when saying

```python
results = ..._fetch_...(...)

if not results:
    do something
```

Because the results still has one item. I'd rather return an empty list. It could lead to some nasty bugs. For example in the above example: When someone does
```python

if results:
  first_item = results[0]
  do_something_with_first_item(first_item)
```

This could have some nasty bugs because the downstream will assume if it results is not empty, then it should be result of successful search.
Also, the return type doesn't match based on the return type-hints?

NISH1001 (changes requested):

akd/tools/search.py
NISH1001: Edit: I see. This is internal method.

But we need to probably remove None items in the actual method:

```python
results_per_query = list(filter(None, results_per_query))
```

NISH1001 (approved):

PR #103, opened 2025-08-05
Add fuzzy matching

iamsims:
- Config use_fuzzy_match and fuzzy_match_threshold

Fuzzy match uses fuzzy token set ratio, in default fuzzy logic is set to false and the fuzzy threshold for match is 87.

### DIFF
--- a/akd/tools/source_validator.py
+++ b/akd/tools/source_validator.py
@@ -7,7 +7,7 @@ against a predefined source whitelist.
 """
 
 from __future__ import annotations
-from rapidfuzz.fuzz import token_set_ratio
+from rapidfuzz.fuzz import token_set_ratio, token_sort_ratio, ratio
 
 import json
 import re
@@ -21,9 +21,17 @@ from akd._base import InputSchema, OutputSchema
 from akd.structures import SearchResultItem
 from akd.tools._base import BaseTool, BaseToolConfig
 from akd.utils import get_akd_root
+from enum import Enum
 
 if TYPE_CHECKING:
     pass
+
+
+
+class FuzzyMatchMethod(str, Enum):
+    token_set = "token_set"
+    token_sort = "token_sort"
+    ratio = "ratio"
 
 
 class SourceInfo(BaseModel):
@@ -115,9 +123,16 @@ class SourceValidatorConfig(BaseToolConfig):
     )
 
     fuzzy_threshold: int = Field(
-        default=87,
-        description="Fuzzy matching threshold (0-100) for journal titles",
+        default=90,
+        description="Fuzzy matching threshold (0-100) for journal titles (default is 90)",
     )
+
+    fuzzy_match_method: FuzzyMatchMethod = Field(
+        default=FuzzyMatchMethod.token_set,
+        description="Fuzzy matching method: token_set (default), token_sort, or ratio",
+    )
+
+
 
 
 class SourceValidator(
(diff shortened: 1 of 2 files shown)


NISH1001:
@iamsims  please  do follow the branch naming convention we have such as `feature/` `bugfix/` `hotfix/` `refactor/` etc

Thanks.

akd/tools/source_validator.py
```diff
@@ -334,18 +345,22 @@ def _validate_against_whitelist(
...
-                # TODO: Could add ISSN matching here if we had ISSN data in whitelist
-
+                
+                 # if (
+                #     whitelisted_title in source_title
+                #     or source_title in whitelisted_title
+                # ):
+                #     # Check if it's a meaningful match (not just common words)
+                #     if len(whitelisted_title) > 10 or len(source_title) > 10:
+                #         return True, category_name, 0.8
```
NISH1001: I think this remove older logic no?

Instead of having destructive change, can we have constructive change. That is: first check if fuzzy is enabled. If not enabled, then use the older logic.  Don't replace current logic because not sure fuzzy will work 100% of the time.

I'd suggest something like

```python
if self.config.use_fuzzy_match:
  ....<your code>
  ...return

continue with whatever logic we have previously. This should fix this comment.
```

akd/tools/source_validator.py
```diff
@@ -334,18 +345,22 @@ def _validate_against_whitelist(
...
+                    
+
+                if self.config.use_fuzzy_match:
+                    fuzzy_score_set = token_set_ratio(source_title, whitelisted_title)
```
NISH1001: Can we have the fuzzy match function name as configurable? That is `fuzzy_fn_name` or something and get the function based on that name. It will give more configurability on what fuzzy matching algo to apply

NISH1001 (changes requested):

NISH1001:
@iamsims ALso let's add source validation test cases as well `tests/tools/test_source_validator` or something like that.

Thanks.

NISH1001:
@iamsims What's the test coverage now?

Could you run `python -m pytest --cov=akd  --cov-report=term-missing tests/` and post the results here of the coverage? Thanks

akd/tools/source_validator.py
```diff
@@ -20,11 +21,19 @@
...
 
 if TYPE_CHECKING:
     pass
 
 
+
+class FuzzyMatchMethod(str, Enum):
+    token_set = "token_set"
+    token_sort = "token_sort"
+    ratio = "ratio"
```
NISH1001: Let's have enum all caps lock like `TOKEN_SET` etc....

```python
TOKEN_SET = "token_set"
...
```

akd/tools/source_validator.py
```diff
@@ -299,6 +325,18 @@ def _parse_crossref_response(
...
+
+    def get_fuzzy_scorer(self):
+        """
+        Returns the appropriate RapidFuzz scorer based on method name.
+        """
+        scorer_map = {
+        "token_set": token_set_ratio,
+        "token_sort": token_sort_ratio,
+        "ratio": ratio,
+        }
```
NISH1001: maybe this could be class-level attribute?

```python

class SourceValidator(...):
  ...
  _scorer_map = dict(token_set = token_set_ratio, ...)
  ...
  def __init__(self, ...):
    ...
```

NISH1001 (changes requested):

NISH1001:
@iamsims is this PR still relevant?

NISH1001:
Closing this as unplanned for now.

PR #119, opened 2025-08-18
Feature/full text search pipeline

iamsims:
This PR introduces the SearchPipeline, a full-text academic search pipeline that integrates search, URL resolution, and scraping into one unified flow. 

##### Input:
SearchToolInputSchema

##### Output:
SearchToolOutputSchema returns a list of enriched SearchResultItem objects.

- The DOI is updated if it is resolved. 
- The url is updated if it is resolved. 
- Additional metadata about the resolution and scraping process is stored in extra.
- Content is populated if scraping is enabled and content is found from the resolved_url/pdf_url/original url. 

Each SearchResultItem looks like:
```

{
  "title": "Paper title",
  "url": "open access / direct PDF URL (if resolved, else original url)",
  "doi": "10.xxxx/xxxxx",
  "content": "Full text (scraped or combined with original)",
  "authors": ["Author1", "Author2"],
  "extra": {
    "scraping_performed": true,
    "full_text_scraped": true,
    "resolver_used": ["DOIResolver", "UnpaywallResolver"],
    "original_url": []
    "scraper_used": "CompositeScraper",
    "scraped_url": "https://..."
  }
}

```

#### Search Integration: 

Uses any configured SearchTool (e.g., SearxNG, Semantic Scholar) to fetch initial SearchResultItem objects.

#### Resolver Layer
Resolves open-access URLs and DOIs using a chain of resolvers:

- PDFUrlResolver
- ArxivResolver
- ADSResolver
- DOIResolver
- CrossRefDoiResolver
- UnpaywallResolver

#### Scraper Layer
Extracts full-text content from resolved URLs using a composite scraper composed of:

- DoclingScraper
- Crawl4AIWebScraper
- SimpleWebScraper
- SimplePDFScraper

#### Pipeline Configuration (SearchPipelineConfig)

- Parallel vs sequential processing
- Max concurrency control
- Scraping timeout handling
- Option to preserve original content alongside scraped content
- Minimum number of successful scrapes required
- Optional fail-fast mode on scraping errors

### Workflow:

1. Fetch results from the underlying search tool.
2. Resolve URLs to open access or direct PDF links.
3. Scrape content from resolved URLs (if enabled).
4. Enhance results by attaching scraped full-text and metadata to SearchResultItem.
5. Return enriched output in SearchToolOutputSchema.

### Usage Snippets 
```

async def get_search_tool():
    _cfg = SearxNGSearchToolConfig(
        engines=["google_scholar", "semanticscholar"],
        max_results=10,
        results_per_page=10)

    _tool = SearxNGSearchTool(config=_cfg)
    return _tool


async def full_pipeline():
    search_tool = await get_search_tool()


    cfg = SearchPipelineConfig( 
        enable_scraping=True,  # Enable content scraping
    )
    pipeline = SearchPipeline(
        search_tool=search_tool,
        config=cfg
    )

    result = await pipeline.arun(
        SearchToolInputSchema(
            queries=["AI in disaster management"], 
            max_results=10,
            results_per_page=10,
        )
    )

    return result



if __name__ == "__main__":
    import asyncio
    result = asyncio.run(full_pipeline())
```


muthukumaranR:
@iamsims can we add description and tests for these, and usage snippets?

akd/tools/resolvers/_base.py
```diff
@@ -0,0 +1,181 @@
...
+    session: Optional[requests.Session] = Field(
+        None,
+        description="Optional session to use for requests",
+    )
+    user_agent: Optional[str] = Field(
+        default="Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
+        "AppleWebKit/537.36 (KHTML, like Gecko) "
+        "Chrome/91.0.4472.124 Safari/537.36",
+        description="Optional user agent to use for requests",
+    )
```
muthukumaranR: This is great - and fixes a lot of denials - may just need to use a true user agent rather than mocking here. i.e. if we are using requests:
```
{
    'User-Agent': 'python-requests/2.22.0', 
    'Accept-Encoding': 'gzip, deflate', 
    'Accept': '*/*', 'Connection': 'keep-alive'
}
```

akd/tools/resolvers/ads.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Returns:
+            Optional[ResolverOutputSchema]: The resolved result with DOI URL or direct PDF URL if found, None otherwise
+        """
+        url = str(params.url)
+        try:
+            self.validate_url(url)
+        except RuntimeError:
+            # Not a valid ADS URL
+            return None
```
muthukumaranR:
```suggestion
        try:
            if not self.validate_url(url)
                return None
        except RuntimeError:
            # handle error with message ...
        
```

muthukumaranR: should we return the url instead of returning None? in case there is a waterfall resolver that can handle url down the line? 
cc: @NISH1001

akd/tools/resolvers/arxiv.py
```diff
@@ -0,0 +1,41 @@
...
+            doi = params.doi or f"10.48550/arXiv.{paper_id}"
+
+            result = ResolverOutputSchema(**params.model_dump())
+            result.resolved_url = pdf_url
+            result.doi = doi
+            result.resolvers.append(self.__class__.__name__)
+            return result
+        except RuntimeError:
+            # Not a valid arxiv URL
+            return None
```
muthukumaranR: same comment as here: 
https://github.com/NASA-IMPACT/accelerated-discovery/pull/119#discussion_r2286626648

akd/tools/resolvers/ads.py
```diff
@@ -0,0 +1,95 @@
...
+        try:
+            # Fetch the ADS page
+            response = self.session.get(url, headers=self.headers)
```
muthukumaranR: may need to use aiohttp.ClientSession for async functions

akd/tools/resolvers/crossref_doi.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    Input schema for CrossRef DOI resolver.
+    Inherits from ResolverInputSchema to ensure compatibility with search results.
+    """
+
+    # makes title  required for CrossRef DOI resolution
+    title: str = Field(
+        ...,
+        description="Title of the article to resolve DOI",
+    )
```
muthukumaranR: I would recommend either enforce title as required across the field, or use a different title field such as crossref_title

iamsims: It is enforced as required

akd/tools/resolvers/ads.py
iamsims: we have that in `ads.py` because 

```
def validate_url(self, url):
        """Check if this URL is from NASA ADS."""
        if "adsabs.harvard.edu" not in str(url):
            raise RuntimeError("Not a valid ADS URL")
        return True
```

so validate url would either give us true or a runtime error, but never false

And when resolve provides us None, we throw error from arun method


Also about returning the runtime error instead of url, it is because when we are working in the waterfall model in `composite.py`



we continue when we get an error 
```

except Exception as e:
                if self.debug:
                    logger.error(f"Error using resolver={resolver_name}: {e}")
                continue
```

akd/tools/resolvers/ads.py
iamsims: has been addressed in the new commit

muthukumaranR (approved):
please make sure the downstream agents that use the `.url` now use `.resolved_url` in case resolving is needed. otherwise, lgtm. thanks for the PR. 

one other issue for a later fix is that we need to use either `httpx` or `aiohttp`, not both... looking at `_base.py`, we need refactor `aiohttp` usage to `httpx`

akd/tools/resolvers/_base.py
```diff
@@ -36,39 +36,27 @@ class ResolverInputSchema(SearchResultItem, InputSchema):
         None,
         description="Query used to obtain the search result",
     )
+        # allow extra fields in input schema
+    class Config:
+        extra = "allow"
```
NISH1001: Convert this to new pydantic v2 style with `model_config` in-built attribute.

Eg:
` {"arbitrary_types_allowed": True}`

akd/tools/resolvers/arxiv.py
```diff
@@ -32,7 +32,10 @@ async def resolve(
             doi = params.doi or f"10.48550/arXiv.{paper_id}"
 
             result = ResolverOutputSchema(**params.model_dump())
-            result.resolved_url = pdf_url
+            if not result.extra:
+                result.extra = {}
```
NISH1001: You can do `result.extra = result.extra or {}`

akd/tools/resolvers/arxiv.py
```diff
@@ -32,7 +32,10 @@ async def resolve(
...
+            result.extra["is_url_resolved"] = True
```
NISH1001, line 35: WIll this throw key error?

akd/tools/resolvers/composite.py
```diff
@@ -38,46 +38,39 @@ async def _arun(
 
         Once the URL is transformed, do NOT replace it again.
         """
-        output = ResolverOutputSchema(**params.model_dump())
+        result = resolver_input = ResolverInputSchema(**params.model_dump())
```
NISH1001: Any info on why this was done? result and resolver_input being assigned to the same object?

akd/tools/resolvers/identity.py
```diff
@@ -21,6 +21,9 @@ async def resolve(
     ) -> Optional[ResolverOutputSchema]:
         """Return the primary URL as-is with preserved metadata"""
         result = ResolverOutputSchema(**params.model_dump())
-        result.resolved_url = params.url
+        
+        if not result.extra:
+            result.extra = {}
```
NISH1001:
```python
result.extra = result.extra or {}
```

akd/tools/resolvers/identity.py
```diff
@@ -21,6 +21,9 @@ async def resolve(
...
+        result.extra["is_url_resolved"] = True
```
NISH1001, line 25: Same: Will there be key error?

akd/tools/resolvers/specialized.py
```diff
@@ -24,7 +24,11 @@ async def resolve(self, params: ResolverInputSchema) -> ResolverOutputSchema | N
         """
         if params.pdf_url:
             result = ResolverOutputSchema(**params.model_dump())
-            result.resolved_url = params.pdf_url
+            if not result.extra:
+                result.extra = {}
```
NISH1001: `result.extra = result.extra or {}`

Also: Is there a way to not do this repeatedly?

I was thinking: in searchresultitem we can do lambda factor for `extra`

eg `strcutures.py`
```python
class SearchResultItem....:
...
...
extra = ...Field(..., default_factor=lambda x: {})
```

This wil solve all the None issue in extra and removes repeated assignment to {}.

akd/tools/search/pipeline.py
```diff
@@ -0,0 +1,400 @@
...
+        ArxivResolver(debug=debug),
+        ADSResolver(debug=debug),
+        DOIResolver(debug=debug),
+        CrossRefDoiResolver(debug=debug),
+        UnpaywallResolver(debug=debug),
+        debug=debug,
+        )
+    
+
+    def default_scraper(self, debug: bool = False) -> ScraperToolBase:
```
NISH1001: mabye converting this to property will be better:

```python

@property
def default_scraper(self) -> ScraperToolBase:
...
...
```

This way we can treat this method as an attribute direclty like `self.scraper = scraper or self.default_scraper`

NISH1001 (changes requested):

akd/tools/resolvers/composite.py
iamsims: Intention in the use of two variables:
- `result`: always holds the last valid ResolverOutputSchema to guarantee we can return something,
            even if all resolvers fail or return None.
- `resolver_input`: working copy passed into each resolver; it is only updated on success. This prevents overwriting `result` with None if a resolver returns no output.

iamsims: I made a change to fix the output type in commit 18bf986564b9c63acda421bc85e9ab556d00b808

akd/tools/resolvers/identity.py
iamsims: No

akd/tools/search/pipeline.py
```diff
@@ -97,24 +97,25 @@ def __init__(self, url: str, message: str):
...
+        DOIResolver(debug=self.debug),
+        CrossRefDoiResolver(debug=self.debug),
+        UnpaywallResolver(debug=self.debug),
+        debug=self.debug,
         )
     
-
-    def default_scraper(self, debug: bool = False) -> ScraperToolBase:
+    @property
+    def _default_scraper(self, debug: bool = False) -> ScraperToolBase:
```
NISH1001: remove the `debug` arg from here because it's not needed because we're already using `self.debug`

NISH1001 (changes requested):

NISH1001 (approved):

PR #183, opened 2025-09-08
Clean the content of crawl4aiscraper

iamsims:
## Changes made 
- Remove the common header footer classes and ids
- Extract the main content if possible

### Stats before this change:
No. of tokens in header:  1014
No. of tokens in footer:  375
No. of tokens in actual content:  13946

### Stats after this change:
No. of tokens in header:  17
No. of tokens in footer: 2
No. of tokens in actual content:  13361

> Note on methodology (heuristic, not exact):
> The numbers above are approximate. I locate the article bounds using simple string anchors ("Research Article" as a start marker and "Download PDF" as an end marker) and then split on whitespace to count “tokens.”
> 

## How to replicate the stats
```

from akd.tools.scrapers import (
    ScraperToolInputSchema,
    Crawl4AIWebScraper,
    DoclingScraper,
)
from akd.tools.scrapers.composite import CompositeScraper

_url = "https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2019WR026574"

_scraper = Crawl4AIWebScraper()

output = await _scraper.arun(ScraperToolInputSchema(url=_url))


article_start = output.content.find("Research Article")
article_end = output.content.find("Download PDF")
end = len(output.content)



text_before = output.content[:article_start].split(" ")
text_after = output.content[article_end:].split(" ")

text_between= output.content[article_start:article_end].split(" ")

print("Headers token: ", len(text_before))
print("Footer token: ", len(text_after))
print("Actual token: ", len(text_between))
```


akd/tools/scrapers/web_scrapers.py
```diff
@@ -162,29 +162,44 @@ async def fetch(self, url: str):
...
+        cleaned_html = self._clean_dom(raw_html)
 
-        soup = BeautifulSoup(html_content, "html.parser")
-        metadata = await self._extract_metadata(
-            soup,
-            Document(html_content),
-            str(params.url),
-        )
+        soup = BeautifulSoup(cleaned_html, "html.parser")
+        markdown = md(cleaned_html).strip()
```
NISH1001: I think markdownify won't work in here. Because crawl4ai itself uses llm to clean and it's better to use crawl4ai result for markdown than markdownify. We should't re-markdownify ourselves because `crawl_result.markdown` already does that which I presume is LLM-based (better)

iamsims: I intercepted to clean the html filtering through tags thats why i didn't used crawl4ai's markdown

NISH1001 (commented):
Won't merge till we fix the comment.

akd/tools/scrapers/web_scrapers.py
NISH1001: THe problem is again: crawl4ai uses LLM-based cleaning which is better in crawl4scraper to be used. We could work with it. Or look into how crawl4ai creates markdown version and see if it's using markdownify or llm?

iamsims:
## Changes
We modify the code to use crawl4ai's default config because markdownify is naive. Instead of cleaning the html and markdownfying we pass the excluded tags and selectors to AsyncWebCrawler

## Outputs
### Stats after this change:
No. of tokens in header: 67
No. of tokens in footer: 185
No. of tokens in actual content: 13805


**Important**
The net noise compared to the current implementation is reduced by 8 times.  For this specific example, the manual cleaning of html has much better noise removal than using crawl4ai -> which is mostly because of the manual process selecting the main-content directly, for which i did not find equivalent implementation in Crawl4AI

akd/tools/scrapers/web_scrapers.py
iamsims: It is addressed in the new commit

iamsims: The exclude tags and exclude selectors function through the following process in Crawl4AIScraper :

- Once content scraping is complete, the specified elements are stripped from the HTML document. Reference: https://github.com/unclecode/crawl4ai/blob/e651e045c44201c83ae68f3ef4858303533f18d9/crawl4ai/content_scraping_strategy.py#L652
- This produces a cleaned_html output from the scraper
- The cleaned_html (when used default markdown generator with default config) is then passed to the markdown generator for conversion to markdown format https://github.com/unclecode/crawl4ai/blob/e651e045c44201c83ae68f3ef4858303533f18d9/crawl4ai/async_webcrawler.py#L561

iamsims:
Test code and Test result file (with urls from agupubs) for evaluation:

[agupubs_scraped_content_comparison.xlsx](https://github.com/user-attachments/files/22781077/agupubs_scraped_content_comparison.xlsx)

[run_crawl4ai.py](https://github.com/user-attachments/files/22781065/run_crawl4ai.py)

akd/tools/scrapers/web_scrapers.py
```diff
@@ -224,6 +224,8 @@ class Crawl4AIScraperConfig(WebScraperToolConfig):
         - Usage examples: Crawl4AIWebScraper class docstring
     """
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
NISH1001, line 227: Is this required now?

akd/tools/scrapers/web_scrapers.py
```diff
@@ -245,6 +247,37 @@ class Crawl4AIScraperConfig(WebScraperToolConfig):
         description="Run browser in headless mode.",
     )
 
+    # filter header and footer by default
+    filter_header_footer: bool = Field(
+        default=True,
+        description="Filter out header and footer elements from content.",
+    )
+
+    excluded_tags : list = Field(
```
NISH1001: proper type hint `list[str]`?

akd/tools/scrapers/web_scrapers.py
```diff
@@ -17,7 +17,7 @@
     WebScraper,
     WebScraperToolConfig,
 )
-
+from markdownify import markdownify as md
```
NISH1001: remove unused imports

NISH1001 (changes requested):

akd/tools/scrapers/web_scrapers.py
iamsims: yes, we require it for the _run_config()'s output type

NISH1001 (approved):

PR #187, opened 2025-09-09
Bugfix: Code Search Result Titles  

pranath-reddy:
### Summary :memo:
This PR fixes an issue where search result titles were shown with placeholders instead of the actual repository names. Additionally, filtered the repo dump and updated the embeddings.  

### Details
- Repository names were added to the repo dump and included in the embedding pipeline.
- Filtered repo dump based on SME curation (Astro) and word count filter on readme text (min 10).
- Switched embedding model to `thenlper/gte-large`
- Updated Google Drive link to the latest repo dump with embeddings.
- Switched reranker model to `cross-encoder/ms-marco-MiniLM-L12-v2`

### Bugfixes :bug:
- Fixed placeholder usage in result titles. 

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval

### DIFF
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -167,7 +167,7 @@ class CombinedCodeSearchToolConfig(CodeSearchToolConfig):
     """
 
     reranker_model_name: str = Field(
-        "cross-encoder/ms-marco-MiniLM-L6-v2",
+        "cross-encoder/ms-marco-MiniLM-L12-v2",
         description="The model to use for reranking the combined results.",
     )
 
@@ -249,17 +249,18 @@ class LocalRepoCodeSearchToolConfig(CodeSearchToolConfig):
     """
 
     data_file: str = str(
-        get_akd_root() / "docs" / "repositories_with_embeddings_v3.csv"
+        get_akd_root() / "docs" / "repositories_with_embeddings_v4.csv"
     )
     google_drive_file_id: str = os.getenv(
         "CODE_SEARCH_FILE_ID",
-        "1QtTKnlQmSFshCvw3cXAHXgMQQ-DMuGq0",
+        "1-3eD0kJFKgsgKhREA4dOW_V2gYfosK9R",
     )
     embedder_type: Literal["sentence-transformers", "openai"] = "sentence-transformers"
     wait_time: int = 1
-    embedding_model_name: str = os.getenv("CODE_SEARCH_MODEL", "all-MiniLM-L6-v2")
+    embedding_model_name: str = os.getenv("CODE_SEARCH_MODEL", "thenlper/gte-large")
     remove_embedding_column: bool = True
     text_column: str = "text"
+    name_column: str = "name"
     desc_column: str = "description"
     embeddings_column: str = "embeddings"
     debug: bool = False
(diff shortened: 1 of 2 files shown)


akd/tools/code_search.py
```diff
@@ -167,7 +167,7 @@ class CombinedCodeSearchToolConfig(CodeSearchToolConfig):
     """
 
     reranker_model_name: str = Field(
-        "cross-encoder/ms-marco-MiniLM-L6-v2",
+        "cross-encoder/ms-marco-MiniLM-L12-v2",
```
NISH1001, line 170: Any reason why the reranker was changed?

akd/tools/code_search.py
```diff
@@ -249,17 +249,18 @@ class LocalRepoCodeSearchToolConfig(CodeSearchToolConfig):
     """
 
     data_file: str = str(
-        get_akd_root() / "docs" / "repositories_with_embeddings_v3.csv"
+        get_akd_root() / "docs" / "repositories_with_embeddings_v4.csv"
```
NISH1001: I think we should probably reuse some env variable to not consistently index the name change? We can default to v4 but sitll get the name from env as well?

akd/tools/code_search.py
```diff
@@ -512,7 +515,7 @@ async def _arun(
 
         formatted_results = [
             SearchResultItem(
-                title=f"GitHub Repository for {query}",
+                title=f"GitHub Repository: {result.pop('name', '')}",
```
NISH1001: Maybe just add the name itself without any prefix. NO need to say "GitHub REpositoiry" becuase it's self-implied. `title = result.pop("name", ...)`

akd/tools/code_search.py
```diff
@@ -742,7 +745,7 @@ async def _arun(
 
         formatted_results = [
             SearchResultItem(
-                title=f"SDE Code Search for {query}",
+                title=f"SDE Code Search Result",
```
NISH1001: Is this required now? we can have a key called `source` inside `extra` field that says which tool was being used? And title itself iwll be just reposotiry name?

tests/code_search_tool_test.py
```diff
@@ -97,7 +97,7 @@ def test_data_file_validation(temp_data_file):
 """Test3: Vector Embedding"""
 
 
-def test_vector_embedding(embedder=Embedder(model_name="all-MiniLM-L6-v2")):
+def test_vector_embedding(embedder=Embedder(model_name="thenlper/gte-large")):
```
NISH1001: Mutable data strcuture as default value is highly discouraged. Maybe use pytest fixture?

NISH1001 (changes requested):

akd/tools/code_search.py
pranath-reddy: Saw slightly better results with L12 compared to L6.

akd/tools/code_search.py
pranath-reddy: Okay, will add the env variable.

akd/tools/code_search.py
pranath-reddy: Yeah, will update it.

akd/tools/code_search.py
pranath-reddy: Not sure SDE has repo titles indexed. Will share the updated dump with them

tests/code_search_tool_test.py
pranath-reddy: Oh okay, will update it

akd/tools/code_search.py
NISH1001: Oh okay. They have the url, right? We can parse the url?

pranath-reddy: Yeah, we can do that. I'll update it

NISH1001 (approved):

PR #189, opened 2025-09-09
Add a new node template for single agent workflow (SingleAgentNodeTemplate)

NISH1001:
**Title:**
`feat: Add SingleAgentNodeTemplate with Guardrails Integration`

-----

**Description:**

### Summary :memo:

This pull request introduces the `SingleAgentNodeTemplate`, a new node template designed to simplify the creation of nodes for single-agent workflows. It automatically extracts the I/O parameters from the provided agent and now includes seamless integration for applying AI safety guardrails directly to the wrapped agent, significantly reducing boilerplate code.

### Details

This change introduces a new class, `SingleAgentNodeTemplate`, and supporting features:

1.  **`SingleAgentNodeTemplate` Class**:
      * Accepts a `BaseAgent` instance and automatically binds its `input_schema` and `output_schema`.
      * The `_execute` logic handles instantiating the agent's input, running the agent, and updating the `node_state` with the output.
2.  **Guardrails Integration**:
      * The constructor now accepts `input_guardrails`, `output_guardrails` (as lists of `RiskDefinition`), and a `guardrails_config` object.
      * This allows for effortless, on-the-fly application of AI safety checks (e.g., for harmful content, jailbreaking) to any agent used within the node.
3.  **Dynamic Guardrail Application**:
      * A new helper function, `apply_guardrails`, has been added to `akd/guardrails.py`. It dynamically applies the guardrails decorator to an agent instance, making the feature easy to use.
4.  **Comprehensive Testing**:
      * A new test file, `tests/nodes/test_single_agent_node_template.py`, provides extensive test coverage for the new template, including basic functionality, various guardrail configurations, and edge cases.
5.  **Minor Improvements**:
      * Removed unused imports from `akd.nodes.states` for better code hygiene.

### Usage

The `SingleAgentNodeTemplate` makes turning an agent into a node simple. You can also add guardrails with just a few extra parameters.

```python
from akd.agents.query import QueryAgent
from akd.nodes.templates import SingleAgentNodeTemplate
from akd.nodes.states import GlobalState, NodeState
from akd.tools.granite_guardian_tool import RiskDefinition

# 1. Instantiate the agent
query_agent = QueryAgent()

# 2. Create the node from the agent. 
#    Guardrails can be added directly during initialization.
query_node = SingleAgentNodeTemplate(
    agent=query_agent,
    node_id="query_node",
    debug=True,
    # Example of adding guardrails:
    input_guardrails=[RiskDefinition.HARM, RiskDefinition.JAILBREAK]
)


# 3. Define the initial state for the node
initial_state = GlobalState(
    node_states={
        "query_node": NodeState(
            inputs={
                "query": "landslide casualties in Nepal last year",
                "num_queries": 10,
            },
        ),
    },
)

# 4. Run the node
final_state = await query_node.arun(initial_state)

# The output of the agent is now stored in the node's state
print(final_state.node_states["query_node"].outputs)
```

### Checks

  - [x] Tested Changes
  - [x] Stakeholder Approval

### DIFF
--- a/akd/guardrails.py
+++ b/akd/guardrails.py
@@ -6,6 +6,7 @@ from loguru import logger
 
 from akd.agents._base import BaseAgent
 from akd.configs.guardrails_config import GuardrailsConfig
+from akd.tools._base import BaseTool
 from akd.tools.granite_guardian_tool import (
     GraniteGuardianInputSchema,
     GraniteGuardianTool,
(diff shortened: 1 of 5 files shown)


akd/guardrails.py
```diff
@@ -240,3 +255,61 @@ def guardrails_validated():
...
 # class MyTool(BaseTool):
 #     pass
+
+
+def apply_guardrails_to_agent(
+    agent: BaseAgent,
+    config: GuardrailsConfig | None = None,
+    input_guardrails: List[RiskDefinition] | None = None,
+    output_guardrails: List[RiskDefinition] | None = None,
+) -> BaseAgent:
```
NISH1001: This probably can apply to BaseTool as well since both follow same semantics. Should we just renmae this to `apply_guardrails_to_things`?

akd/nodes/templates.py
```diff
@@ -231,3 +238,109 @@ async def _execute(
         node_state.steps.update(sup_out.steps)
 
         return node_state
+
+
+class SingleAgentNodeTemplate(AbstractNodeTemplate):
```
NISH1001, line 243: I am restrciting this to agent level...but the sematnics will work for Tool as well since `BaseAgent` and `BaseTool` follow same IO semantics. But for us, let's just use `agent-level` node as that makes sense. Or should we enforce both tools and agents?

akd/guardrails.py
muthukumaranR: I agree. I have run the wrapper on tools and seemed to work too... I think just apply_guardrails(AbstractBase) could work as well.

muthukumaranR:
```python
  def apply_guardrails(
      component: Union[BaseAgent, BaseTool],
      config: GuardrailsConfig | None = None,
      input_guardrails: List[RiskDefinition] | None = None,
      output_guardrails: List[RiskDefinition] | None = None,
  ) -> Union[BaseAgent, BaseTool]:
  ```
  
  something like this?

File: akd/nodes/templates.py (hunk 2 above)
Comment (muthukumaranR) [reply akd/nodes/templates.py:243]: I would recommend your same restriction - keep it to agent level. since the node templates are agent-focused anyway. if the need arises that a tool is complex enough that needs it's own state management and checkpointing, then we can revisit this. for now we can treat tools as integrated components within agents.

File: akd/guardrails.py [hunk 3]
```diff
@@ -240,3 +255,61 @@ def guardrails_validated():
...
+    has_config = config is not None
+
+    if has_input_guardrails or has_output_guardrails or has_config:
+        # Apply the decorator to create a guarded agent class
+        logger.info(f"Applying guardrails to agent {agent_name}")
+        GuardedAgentClass = add_guardrails(
+            input_guardrails=input_guardrails,
+            output_guardrails=output_guardrails,
+            config=config,
+        )(agent.__class__)
```
muthukumaranR: I think we may want to revisit how we are dynamically creating class object here, as well as in `def decorator(cls):` - could be problematic..

This will create the `GuardedAgentClass` in memory with no way of terminating it (or it being GC'd) unless the process itself is terminated. May be okay for short running processes but could be an issue for long running processes.

akd/guardrails.py
```diff
@@ -240,3 +255,61 @@ def guardrails_validated():
...
+
+        # Create new guarded agent instance preserving original state
+        guarded_agent = GuardedAgentClass.__new__(GuardedAgentClass)
+        guarded_agent.__dict__.update(agent.__dict__)
+
+        # Initialize the guardrails system
+        GuardedAgentClass.__init__(guarded_agent)
```
muthukumaranR: Should we do a safer copy here instead? maybe use `deepcopy()`?

muthukumaranR (commented):
Tested on my end for functionality, all tests pass. I've added review comments - feel free to resolve and merge.

akd/guardrails.py
NISH1001: I had tried on `apply_guardrails` function to not mutate the incoming agent, but seems like copy is not possible because of some `rthread lock` issue which as to do with HTTPClient or something.

Eg:
```python

agent = QueryAgent()

agent_cp = copy.deepcopy(agent) # will throw error
```

```
TypeError: cannot pickle '_thread.RLock' object
```

NISH1001: Re: On the guarded_agent, it's already a copy no? Since we're basically creating a new class at runtime which is scoped within the decorator itself.

akd/nodes/templates.py
NISH1001: Agreed.

akd/guardrails.py
NISH1001: It will be scoped within the decorator itself. Since, guardrailed function will be running till the workflow finishes or backend terminates the session, it will be there. But I don't think it will have any substantial toll on memory as we're just creating classes adhoc.

NISH1001: But, yes. There has to be other better way to do it. We can look into it later.

akd/guardrails.py
NISH1001: I can lookinto the copy in another PR. Seems like the culprlit is `openai.AsyncOpenAIClient` in the base class and we might have to implement `__deepcopy__` method there. I will create another PR for that

muthukumaranR (approved):

PR #198, opened 2025-09-10
Agent Registry Foundation

muthukumaranR:
## Summary

Adds Agent Registry system for auto-discovery and configuration of AKD agents.

## Key Features

- **Auto-discovery**: Automatically finds and registers agents from known modules
- **JSON persistence**: Generates `akd/mapping/agent_registry.json` with agent definitions
- **Selective loading**: Configure which agents to load via settings
- **Schema extraction**: Captures input/output schemas from agent classes

## Generate Registry JSON

The registry auto-generates when the JSON file is missing or empty:

```python
from akd.planner.registry import AgentRegistry

# Auto-discovers all available agents
registry = AgentRegistry()

# Or configure specific agents only
from akd.planner.config import AgentRegistryConfig
config = AgentRegistryConfig(enabled_agents=["query", "extraction"])
registry = AgentRegistry(config)
```

## Configuration

Configure via `AgentRegistryConfig`:

```python
config = AgentRegistryConfig(
    registry_path="custom/path/registry.json",  # Default: akd/mapping/agent_registry.json
    auto_discover=True,                         # Auto-find agents when missing
    enabled_agents=["query", "extraction"],    # Limit to specific agents ([] = all)
    validate_schemas=True                       # Ensure agents have valid schemas
)
```

## How It Works

1. **Discovery**: Scans `KNOWN_AGENTS` list for agent modules
2. **Schema extraction**: Uses `model_json_schema()` to capture agent input/output definitions
3. **JSON generation**: Creates structured registry file with agent metadata
4. **Loading**: Provides methods to query agents by ID, tag, or enabled status

The generated JSON contains agent definitions with schemas, descriptions, tags.

### DIFF
--- a/akd/planner/config.py
+++ b/akd/planner/config.py
@@ -1,0 +1,32 @@
+"""
+Configuration for the AKD Planner module.
+"""
+
+from typing import List
+from pydantic import Field
+
+from akd._base import BaseConfig
+
+
+class AgentRegistryConfig(BaseConfig):
+    """Configuration for the Agent Registry."""
+    
+    registry_path: str = Field(
+        default="akd/mapping/agent_registry.json",
+        description="Path to the agent registry JSON file"
+    )
+    
+    auto_discover: bool = Field(
+        default=True, 
+        description="Auto-populate registry if empty or missing"
+    )
+    
+    enabled_agents: List[str] = Field(
+        default_factory=list,
+        description="List of enabled agent IDs. Empty list enables all agents"
+    )
+    
+    validate_schemas: bool = Field(
+        default=True,
+        description="Validate agent schemas during discovery"
+    )
--- a/akd/planner/config.py
+++ b/akd/planner/config.py
@@ -2,17 +2,17 @@
 Configuration for the AKD Planner module.
 """
 
-from typing import List
 from pydantic import Field
 
 from akd._base import BaseConfig
+from akd.utils import get_akd_root
 
 
 class AgentRegistryConfig(BaseConfig):
     """Configuration for the Agent Registry."""
     
     registry_path: str = Field(
-        default="akd/mapping/agent_registry.json",
+        default=str(get_akd_root() / "akd" / "mapping" / "agent_registry.json"),
         description="Path to the agent registry JSON file"
     )
     
@@ -21,7 +21,7 @@ class AgentRegistryConfig(BaseConfig):
         description="Auto-populate registry if empty or missing"
     )
     
-    enabled_agents: List[str] = Field(
+    enabled_agents: list[str] = Field(
         default_factory=list,
         description="List of enabled agent IDs. Empty list enables all agents"
     )
--- a/akd/planner/registry.py
+++ b/akd/planner/registry.py
@@ -1,0 +1,286 @@
+"""
+Agent Registry for the AKD Planner module.
+
+This module provides agent registration and discovery capabilities for the AKD framework.
+"""
+
+import importlib
+import json
+import os
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from datetime import datetime, timezone
+
+from loguru import logger
+from pydantic import BaseModel, Field
+
+from akd._base import IOSchema
+from .config import AgentRegistryConfig
+
+
+class FieldDefinition(BaseModel):
+    """Individual field definition for agent schemas."""
+    
+    name: str = Field(description="Field name")
+    type: str = Field(description="Field type")
+    description: str = Field(description="Field description")
+    required: bool = Field(default=True, description="Whether field is required")
+    default: Union[str, int, float, bool, List[Any], None] = Field(default=None, description="Default value if any")
+    items_type: Optional[str] = Field(default=None, description="Array item type")
+
+
+class AgentSchemaDefinition(BaseModel):
+    """Schema definition for agent inputs/outputs."""
+    
+    fields: List[FieldDefinition] = Field(
+        default_factory=list,
+        description="List of field definitions"
+    )
+
+
+class AgentEntry(BaseModel):
+    """Registry entry for a single agent."""
+    
+    agent_id: str = Field(description="Unique agent identifier")
+    name: str = Field(description="Human-readable agent name")
+    description: str = Field(description="Agent description")
+    agent_class: str = Field(description="Python import path to agent class")
+    enabled: bool = Field(default=True, description="Whether agent is enabled")
+    input_schema: AgentSchemaDefinition = Field(description="Input schema definition")
+    output_schema: AgentSchemaDefinition = Field(description="Output schema definition")
+    tags: List[str] = Field(default_factory=list, description="Agent tags")
+    use_cases: List[str] = Field(default_factory=list, description="Agent use cases")
+    dependencies: List[str] = Field(default_factory=list, description="Agent dependencies")
+
+
+class AgentRegistryData(BaseModel):
+    """Top-level registry data structure."""
+    
+    version: str = Field(default="1.0.0", description="Registry format version")
+    created_at: str = Field(default_factory=lambda: datetime.now(timezone.utc).isoformat())
+    updated_at: str = Field(default_factory=lambda: datetime.now(timezone.utc).isoformat())
+    agents: Dict[str, AgentEntry] = Field(default_factory=dict, description="Agent entries")
+
+
+class AgentRegistry:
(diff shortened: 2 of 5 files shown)


akd/planner/registry.py
```diff
@@ -0,0 +1,286 @@
...
+
+from akd._base import IOSchema
+from .config import AgentRegistryConfig
+
+
+class FieldDefinition(BaseModel):
+    """Individual field definition for agent schemas."""
+    
+    name: str = Field(description="Field name")
+    type: str = Field(description="Field type")
```
NISH1001: Are these required fields? Then we could use elipses `...` in the `Field(..., descriptoin="...")`

akd/planner/registry.py
```diff
@@ -0,0 +1,286 @@
...
+    description: str = Field(description="Field description")
+    required: bool = Field(default=True, description="Whether field is required")
+    default: Union[str, int, float, bool, List[Any], None] = Field(default=None, description="Default value if any")
```
NISH1001: Nitpick:
Maybe let's start to follow new python 3.12+ typing style with 
```
default: str | int | float | bool | list[Any] | None = ...
```

akd/planner/config.py
```diff
@@ -0,0 +1,32 @@
...
+from pydantic import Field
+
+from akd._base import BaseConfig
+
+
+class AgentRegistryConfig(BaseConfig):
+    """Configuration for the Agent Registry."""
+    
+    registry_path: str = Field(
+        default="akd/mapping/agent_registry.json",
```
NISH1001: use probably adk root-based path

```python
from akd.utils import get_akd_root
```
and take it from there? it iwll be better...We've done it on several places. [eg](https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/akd/tools/code_search.py#L251)

akd/planner/registry.py
```diff
@@ -0,0 +1,286 @@
...
+        ("followup_query", "akd.agents.query", "FollowUpQueryAgent"), 
+        ("extraction", "akd.agents.extraction", "EstimationExtractionAgent"),
+        ("relevancy", "akd.agents.relevancy", "MultiRubricRelevancyAgent"),
+        ("intent", "akd.agents.intents", "IntentAgent"),
+        ("controlled_search", "akd.agents.search.controlled", "ControlledSearchAgent"),
+        ("deep_search", "akd.agents.search.deep_search", "DeepLitSearchAgent"),
+        ("gap_analysis", "akd.agents.gap_analysis.gap_analysis", "GapAgent"),
+        ("storm", "akd.agents.storm.storm", "StormAgent"),
+        ("aspect_search", "akd.agents.search.aspect_search.aspect_search", "AspectSearchAgent"),
+    ]
```
NISH1001: Should we do this as mapping of Dictionary to tuple?

```python
KNOWN_AGENTs  = {
"query" : (akd.agents.query, QueryAgent)
}
```

etc?

akd/planner/registry.py
```diff
@@ -0,0 +1,286 @@
...
+# Global registry instance
+_registry: Optional[AgentRegistry] = None
+
+
+def get_agent_registry(config: Optional[AgentRegistryConfig] = None) -> AgentRegistry:
+    """Get the global agent registry instance."""
+    global _registry
+    if _registry is None:
+        _registry = AgentRegistry(config)
+    return _registry
```
NISH1001: This exhibits singleton, right? or maybe we can inherently make the registry singleton?

```python
class AgentRegistry:

_registry = None

def __new__(self....)->
... if _regsitry is None: intialize new
else just use _registry?
```

what do you think?

Gemini-recommended snippet:
```python
class AgentRegistry:
    # 1. A class attribute to hold the single instance
    _instance = None

    # 2. Override the __new__ method
    def __new__(cls, *args, **kwargs):
        # If the single instance doesn't exist yet...
        if cls._instance is None:
            # ...create it by calling the parent class's __new__ method
            # and store it in our class attribute.
            cls._instance = super().__new__(cls)
        # Always return the single instance
        return cls._instance

    # The rest of the class ( __init__, _load_or_discover, etc.) stays the same
    def __init__(self, config: Optional[AgentRegistryConfig] = None):
        """Initialize the agent registry."""
        # Note: __init__ will run every time you call AgentRegistry(),
        # so you might need to add a check to prevent re-initialization.
        # For example: if hasattr(self, 'config'): return
        self.config = config or AgentRegistryConfig()
        self.registry_data: AgentRegistryData = AgentRegistryData()
        self._load_or_discover()

    # ... other methods ...
```

This way we can always reuse the same object and minimizes potential side-effects of registry mutation. IF mutation / registry is needed to be updated at runtime, we can allow through some methods and it will be scoped to everywhere since only single instance can be created/reused.

NISH1001 (commented):
singleton comments

akd/planner/registry.py
muthukumaranR: thanks for catching this - updated.

akd/planner/registry.py
muthukumaranR: changed to use the python 3.12 types here: [822ce86](https://github.com/NASA-IMPACT/accelerated-discovery/pull/198/commits/822ce86c68bc6e47afcc48c9ecd152583a14de9f)

akd/planner/registry.py
muthukumaranR: good suggestion - [99cc1c2](https://github.com/NASA-IMPACT/accelerated-discovery/pull/198/commits/99cc1c2a7367543adf7ee05d5831422593e5ea01)

  Replaced the global _registry variable with singleton implementation:
  - Added _instance and _initialized class attributes
  - Implemented __new__ to return single instance
  - Modified __init__ to initialize only once

akd/planner/registry.py
muthukumaranR: [822ce86](https://github.com/NASA-IMPACT/accelerated-discovery/pull/198/commits/822ce86c68bc6e47afcc48c9ecd152583a14de9f)

akd/planner/registry.py
muthukumaranR: updated to use 3.12 python types

akd/planner/config.py
muthukumaranR: fixed here [d71e6d2](https://github.com/NASA-IMPACT/accelerated-discovery/pull/198/commits/d71e6d201f70b93db46ce597c7f32cb2fbaf6961)

akd/planner/registry.py
muthukumaranR: 734ac77

akd/planner/registry.py
```diff
@@ -67,29 +68,43 @@ class AgentRegistry:
...
-        """Initialize the agent registry."""
-        self.config = config or AgentRegistryConfig()
-        self.registry_data: AgentRegistryData = AgentRegistryData()
-        self._load_or_discover()
+        """Initialize the agent registry (only once due to singleton pattern)."""
+        if not self._initialized:
+            self.config = config or AgentRegistryConfig()
+            self.registry_data: AgentRegistryData = AgentRegistryData()
+            self._load_or_discover()
+            AgentRegistry._initialized = True
```
NISH1001: Nice. This makes it cleaner to put singleton.

NISH1001 (approved):

NISH1001:
@muthukumaranR there seems to be some tests failing. ON dev we have 3 failed tests. Here we have 6. And they are related to i think `mapping`?

log here:
https://github.com/NASA-IMPACT/accelerated-discovery/actions/runs/17645221017/job/50141434205

NISH1001 (approved):

muthukumaranR:
Okay - I have fixed the tests failing. there are still some tests failing but they are unrelated to the updates in this PR: they are related to `search/deep_search.py` and the output format - I will fix them in a different PR. I will merge this PR now.


```
FAILED tests/agents/search/test_deep_search.py::TestDeepLitSearchAgentIntegration::test_iterative_research_with_quality_threshold - StopAsyncIteration
FAILED tests/agents/search/test_deep_search.py::TestDeepLitSearchAgentErrorHandling::test_component_failure_graceful_degradation - Exception: Triage failed
FAILED tests/mapping/test_mappers.py::TestRealAgentMappings::test_full_agent_pipeline_mapping - pydantic_core._pydantic_core.ValidationError: 2 validation errors for LitSearchAgentOutputSchema
results.0
```

PR #200, opened 2025-09-15
Add Code Repository Search Agent

pranath-reddy:
### Summary :memo:
This PR adds an initial implementation of the `CodeSearchAgent`, a configurable agent that performs iterative code repository search by reformulating queries. It coordinates multiple sub agents and tools to enable a reasoning-driven search. It takes a user query, reformulates it using `QueryAgent`, performs code search via a pluggable `CodeSearchTool`, evaluates results using a rubric-based `MultiRubricRelevancyAgent`, and refines queries through a `FollowUpQueryAgent`. The process continues iteratively until sufficient alignment is reached or the iteration limit is hit.

### Details
- Introduced `CodeSearchAgent` under `akd.agents.search.code_search` to support iterative, LLM-guided code repository search.
- Inherits from `ControlledSearchAgent` to leverage existing search logic and iteration control.
- Fully modular via dependency injection:
  - `QueryAgent` for reformulating the initial query.
  - `CodeSearchTool` (e.g., `CombinedCodeSearchTool`, `LocalRepoCodeSearchTool`) for retrieving search results.
  - `MultiRubricRelevancyAgent` for evaluating alignment of search results with query intent.
  - `FollowUpQueryAgent` for refining queries based on content and feedback.
- Extends `ControlledSearchAgentConfig` with code-specific parameters:
  - Embedding model configuration for local search
  - SDE search API configuration
  - Tool selection flags for local/SDE search components
  - Model selection for sub-agents
- Added `is_server_available` under akd.utils for URL format validation and reachability testing

### Usage
```python
from akd.agents.search import CodeSearchAgent, CodeSearchAgentConfig
from akd.agents.search import LitSearchAgentInputSchema

# Configure and initialize the agent
cfg = CodeSearchAgentConfig()
agent = CodeSearchAgent(config=cfg)

# Run the search
result = await agent.arun(
    LitSearchAgentInputSchema(
        query="landslide nepal", 
        max_results=10
    )
)
```
```python
from akd.utils import is_server_available

url = "https://www.google.com"
is_server_available(url)
```

### Checks
- [x] Tested Changes
- [x] Stakeholder Approval


akd/agents/search/code_search.py
```diff
@@ -0,0 +1,187 @@
...
+        default=None, description="Path to local repository data file"
+    )
+    google_drive_file_id: Optional[str] = Field(
+        default="15kxTyLeBCPL82WjMTyDytcXag85vglCP",
+        description="Google Drive file ID for repository database, uses gte-large embeddings",
+    )
+
+    # SDE search configuration
+    sde_base_url: str = Field(
+        default="https://d2kqty7z3q8ugg.cloudfront.net/api/code/search",
```
NISH1001: can we by default get from os env `SDE_BASE_URL` and default to this URL?

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,187 @@
...
+        description="SDE search API base URL",
+    )
+    sde_search_type: str = Field(
```
NISH1001: use `Literal["vector", "hybrid", "keyword"]` to restrict the values

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,187 @@
...
+        query_agent: Optional[QueryAgent] = None,
+        followup_query_agent: Optional[FollowUpQueryAgent] = None,
+        relevancy_agent: Optional[MultiRubricRelevancyAgent] = None,
+        debug: Optional[bool] = None,
+    ):
+        """Initialize CodeSearchAgent with custom components."""
+
+        self.config = config or CodeSearchAgentConfig()
+        if debug is not None:
+            self.config.debug = debug
```
NISH1001: Is this required?

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+
+    def __init__(
+        self,
+        config: Optional[CodeSearchAgentConfig] = None,
+        search_tool: Optional[CombinedCodeSearchTool] = None,
+        query_agent: Optional[QueryAgent] = None,
+        followup_query_agent: Optional[FollowUpQueryAgent] = None,
+        relevancy_agent: Optional[MultiRubricRelevancyAgent] = None,
+        debug: Optional[bool] = None,
```
NISH1001: `debug: bool = False` is better way

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,187 @@
...
+class CodeSearchAgent:
+    """
+    Wrapper for code repository search using ControlledSearchAgent.
+    """
+
+    def __init__(
+        self,
+        config: Optional[CodeSearchAgentConfig] = None,
+        search_tool: Optional[CombinedCodeSearchTool] = None,
+        query_agent: Optional[QueryAgent] = None,
```
NISH1001: Let's use new python 3.12+ type hint style
```python
config: CodeSearchAgentConifg | None = None
search_tool: CombinedSearchTool | None = None
...
...
```

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,187 @@
...
+
+class CodeSearchAgent:
+    """
+    Wrapper for code repository search using ControlledSearchAgent.
+    """
+
+    def __init__(
+        self,
+        config: Optional[CodeSearchAgentConfig] = None,
+        search_tool: Optional[CombinedCodeSearchTool] = None,
```
NISH1001: The type for this should be `CodeSearchTool` no? Which wil give same semantics no matter what input tools are.

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,187 @@
...
+        # Setup SDE search tool
+        if self.config.use_sde_search:
+            sde_config = SDECodeSearchToolConfig(
+                base_url=self.config.sde_base_url,
+                debug=self.config.debug,
+                search_mode=self.config.sde_search_type,
+                page_size=self.config.sde_page_size,
+            )
+            sde_tool = SDECodeSearchTool(config=sde_config)
+            tools.append(sde_tool)
```
NISH1001: Do we need this?

I was thinking, since the incoming `search_tool` will get passed something or None.
```python
search_tool = search_tool or CombinedSearchTool([LocalTool(...), SDE...()])
```

This way, if we want to enable what to have for SDE etc, we can depednecny-inject direclty.

Maybe, for this method, better way could be to have a classmethod-initilized thing:

```python
class CodeSearchAgent(...):
..
...

@classmethod
def from_local_sde_config(cls, config_agent: CodeSearchAgentConfig | None = None, local_search_config...., sde_search_config,....) -> CodeSearchAgent:
  # initialize combined tool here direclty and pass to the constructor
  tools = []
  if local_search_config:
    tools.append(Local...(local_search_config))...
  elif sde_search_config:
    ....
  tool = CombinedSearchTool(...)
  return cls(config=agent_config, tool=tool, ...)
```

Something like this. This iwll help to get more ctonrol on how to initialize and segregate responsibility

NISH1001 (changes requested):

akd/agents/search/code_search.py
pranath-reddy: I am creating the tools internally since I didn't know whether having external tool configs/definitions would be compatible with the node template.

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,248 @@
...
+        """Initialize CodeSearchAgent with custom components."""
+
+        self.config = config or CodeSearchAgentConfig()
+        self.config.debug = debug
+
+        # Setup search tool, prefer to use the provided search tool
+        if search_tool is not None:
+            self.search_tool = search_tool
+        else:
+            self.search_tool = self._setup_search_tool()
```
NISH1001:
```python
self.search_tool = search_tool or self._setup_search_tool()
```

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,248 @@
...
+    # Tool selection
+    use_local_search: bool = Field(
+        default=True, description="Enable local repository search"
+    )
+    use_sde_search: bool = Field(
+        default=True, description="Enable SDE repository search"
+    )
+
+
+class CodeSearchAgent:
```
NISH1001: Let's inherit from `ControlledSearchAgent` and rework some logic accordingly as most of the logic we can reuse from the parent class. ANd only change the logic if needed?

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+        # Setup agents
+        self.query_agent = query_agent or self._setup_query_agent()
+        self.followup_query_agent = (
+            followup_query_agent or self._setup_followup_query_agent()
+        )
+        self.relevancy_agent = relevancy_agent or self._setup_relevancy_agent()
+
+        # Create the underlying ControlledSearchAgent
+        self._controlled_agent = ControlledSearchAgent(
```
NISH1001: if we inherit from `ControlledSEarchAgent`, we don't need to hae this as a sub agent as the whole agent will be this agent.

NISH1001 (changes requested):
Pass I

NISH1001:
@pranath-reddy could you pull in the latest develop branch. There, the failed tests are 0 so far. So need to reflect the test coverage here as well. Thanks.

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,227 @@
...
+        description="Re-ranker model for combined search",
+    )
+
+    # Local search configuration
+    embedding_model_name: str = Field(
+        default="thenlper/gte-large", description="Embedding model for local search"
+    )
+    data_file: str | None = Field(
+        default=None, description="Path to local repository data file"
+    )
```
NISH1001, line 48: What data file is this?

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,227 @@
...
+                    if self.config.debug:
+                        logger.info(
+                            f"[CodeSearchAgent] Verified reachability of SDE URL: {url}"
+                        )
+                except requests.RequestException as e:
+                    reachable = False
+                    if self.config.debug:
+                        logger.warning(
+                            f"[CodeSearchAgent] Could not reach SDE URL '{url}'. Skipping SDE tool. Reason: {e}"
+                        )
```
NISH1001: Can we maybe move this test to `akd.utils.py`? That is: maybe `is_server_available(url...) -> bool` or something and reuse that? It could be handy for some other cases also within akd.

```python
if self.config.use_sde_search:
  reachable = is_server_available(sde_base_url)
```

```python
def is_server_available(url: str | HttpURL) -> bool:
  # test for url validity first
  # then requests.head(...)
  # if any exceptions or non 2xx status, return False...
```

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,227 @@
...
+                        page_size=self.config.sde_page_size,
+                    )
+                    sde_tool = SDECodeSearchTool(config=sde_config)
+                    tools.append(sde_tool)
+
+            except Exception as e:
+                if self.config.debug:
+                    logger.warning(
+                        f"[CodeSearchAgent] SDECodeSearchTool unavailable; continuing without it. Reason: {e}"
+                    )
```
NISH1001: If we have the `is_server_available` util, then i think this method will look less complicated and help flatten the nested try catch as well.

NISH1001 (changes requested):

akd/agents/search/code_search.py
pranath-reddy: It's for the repo dump

akd/agents/search/code_search.py
pranath-reddy: okay, will add to utils

pranath-reddy:
> @pranath-reddy could you pull in the latest develop branch. There, the failed tests are 0 so far. So need to reflect the test coverage here as well. Thanks.

One failed test. Looks like there's a missing field in the output of code search agent. Will fix it. Was LitSearchAgentOutputSchema updated?

NISH1001:
> > @pranath-reddy could you pull in the latest develop branch. There, the failed tests are 0 so far. So need to reflect the test coverage here as well. Thanks.
> 
> One failed test. Looks like there's a missing field in the output of code search agent. Will fix it. Was LitSearchAgentOutputSchema updated?

Yes. I presume we added "extra" field in it. but that should be default to empty dict.

akd/agents/search/code_search.py
```diff
@@ -0,0 +1,227 @@
...
+
+                if local_config is not None:
+                    local_tool = LocalRepoCodeSearchTool(config=local_config)
+                    tools.append(local_tool)
+
+            except Exception as e:
+                if self.config.debug:
+                    logger.warning(
+                        f"[CodeSearchAgent] LocalRepoCodeSearchTool unavailable; continuing without it. Reason: {e}"
+                    )
```
NISH1001: also maybe segregate the logic to an ewm ethod like `_setup_local_tool(...) -> LocalSearchTool | None` or something?

This iwll clean the setup method here?

NISH1001 (changes requested):

NISH1001:
@pranath-reddy Tangentially related to this PR:
Can we also move the code search tool to the `akd.tools.search` name space? And throw deprecation warning on existing code search tool namespace? We can have that in another PR.

Something like:
in `akd.tools.code_searc.py`
```python
from warnings import warn

warn('The `akd.tools.code_search` modeule is deprecated and moved to `akd.tools.search.code_search`', DeprecationWarning, stacklevel=2)
```

NISH1001 (approved):

PR #211, opened 2025-09-22
Feat: Make Base Agents Stateless by Default

NISH1001:
### Summary :memo:

This PR refactors the base agent classes (`LangBaseAgent` and `InstructorBaseAgent`) to be **stateless by default**. A new `stateless` configuration flag has been introduced, which controls whether the agent maintains conversation history across multiple `arun` calls. When `stateless=True` (the new default), the agent's memory is not updated, making each call independent.

### Details

1.  **Added `stateless` Config Flag**: A new boolean field `stateless` was added to `BaseAgentConfig`, which defaults to `True`.
2.  **Refactored `_arun` Method**: The `_arun` method in the base classes now creates a temporary message history for each call when in stateless mode. The agent's persistent `self.memory` is only updated if `stateless` is set to `False`.
3.  **Updated Unit Tests**: Tests were updated to reflect the new default behavior. New tests were added to explicitly verify that memory remains empty in stateless mode and is correctly updated in stateful mode (`stateless=False`).

## Usage

The following example demonstrates how to use the agent in its default stateless mode. After the `arun` call, the agent's internal memory will remain empty. To persist the conversation history, set `stateless=False` in the configuration.

```python
from akd.agents.query import (
    QueryAgent,
    QueryAgentInputSchema,
    QueryAgentOutputSchema,
)

from akd.agents._base import BaseAgentConfig

# The 'stateless=True' flag ensures memory is not stored after the call.
# This is the default behavior.
_cfg = BaseAgentConfig(
    model_name="gpt-4o-mini", temperature=0.5, stateless=True
)

agent = QueryAgent(_cfg)

_query = "what is the meaning of life"

_output = await agent.arun(QueryAgentInputSchema(query=_query, num_queries=3))

print("Agent Output:")
print(_output)

# The agent's memory will be empty because it ran in stateless mode.
print("\nAgent Memory:")
print(agent.memory)
```

### Checks

  - [x] Tested Changes (local coverage 77%, no failure, 315 passed)
  - [ ] Stakeholder Approval

### DIFF
--- a/akd/agents/_base.py
+++ b/akd/agents/_base.py
@@ -23,6 +23,10 @@ class BaseAgentConfig(BaseConfig):
     model_name: str | None = Field(default=CONFIG.model_config_settings.model_name)
     temperature: float = 0.0
     system_prompt: str | None = Field(default=DEFAULT_SYSTEM_PROMPT)
+    track_history: bool = Field(
+        default=False,
+        description="Whether to track conversation history",
+    )
 
 
 class BaseAgent[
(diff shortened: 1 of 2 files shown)


muthukumaranR (approved):
thanks for the change. lgtm

PR #217, opened 2025-09-25
Enhance System Prompt with Input Schema Descriptions

NISH1001:
### Summary :memo:
This PR introduces a new feature that enhances the agent's system prompt by automatically including descriptions of the `input_schema` fields. This is controlled by a new `input_hints` boolean flag in the `BaseAgentConfig`. When enabled, this "input hinting" provides the LLM with clearer context about the expected input, aiming to improve the accuracy and reliability of its responses.

Fields without descriptions are skipped automatically if this hinting is enabled.

### Details

1. **Added `input_hints` Flag**: A new boolean field `input_hints` was added to `BaseAgentConfig`, defaulting to `False`.
2. **Enhanced System Prompt**: Implemented a new `_system_prompt` property in `BaseAgent` that dynamically appends the formatted input schema information to the user-defined `system_prompt` if `input_hints` is `True`.
3. **Schema Utility**: Created a new utility function, `akd.utils.get_model_fields(...)`, to reliably extract field names, descriptions, and other metadata from a Pydantic model's schema (fields without descriptoins are skipped).
4. **Agent Integration**: Updated both `LangBaseAgent` and `InstructorBaseAgent` to use the new `_system_prompt` property, ensuring consistent behavior across all agent types.
5. **Added Unit Tests**: Wrote comprehensive tests for both agent base classes to verify the functionality of the `input_hints` flag, including scenarios where it is enabled, disabled, and when no input schema is present.

## Bugfix
- Fixed a bug in `akd.agents._base.BaseAgentConfig.temperature` value where it was defaulting to 0.0 hard-coded. Now, it's changed to reflect project global config. This way we can control globally this parameter if we want. Default value in the project config is 0.0 as well.
  - This fixes few things when we are running everything on gpt-5 models whose temperature is only 1.0 in the API. So, changing globally this model without this fix will throw error. Now the bug is fixed.

### Example

```python
from akd._base import InputSchema, OutputSchema
class HaikuAgentInputSchema(InputSchema):
    """
    Input schema that takes in query
    """

    query: str = Field(
        ...,
        description="Query to the haiku agent.",
    )

    topic_steer: str = Field(
        default="5-3-5 format",
        description="Steer in 5-3-5 syllable as possible. Also use sanksrit words in the middle line"
    )


class HaikuAgentOutputSchema(OutputSchema):
    """
    Generate a haiku from given query
    """

    haiku: str = Field(..., description="Haiku content")

class HaikuAgent(LiteLLMInstructorBaseAgent):
    input_schema = HaikuAgentInputSchema
    output_schema = HaikuAgentOutputSchema

config = BaseAgentConfig(input_hints=True, system_prompt="You're a haiku master.")

agent = HaikuAgent(config)

print(agent._system_prompt)
```

```
You're a haiku master.

INPUT FIELD DESCRIPTIONS:
- **query**: Query to the haiku agent.
- **topic_steer**: Steer in 5-3-5 syllable as possible. Also use sanksrit words in the middle line
```

### Checks
- [x] Tested Changes (356 passed, 0 failed, 79% coverage)
- [ ] Stakeholder Approval

### DIFF
--- a/akd/agents/_base.py
+++ b/akd/agents/_base.py
@@ -11,7 +11,7 @@ from langchain_openai import ChatOpenAI
 from litellm import acompletion, get_model_info
 from litellm.utils import trim_messages
 from loguru import logger
-from pydantic import AnyUrl, BaseModel, Field, model_validator
+from pydantic import AnyUrl, BaseModel, Field, create_model, model_validator
 
 from akd._base import AbstractBase, BaseConfig, InputSchema, OutputSchema
 from akd.configs.project import CONFIG
@@ -34,6 +34,10 @@ class BaseAgentConfig(BaseConfig):
     stateless: bool = Field(
         default=True,
         description="Whether to maintain conversation history/state",
+    )
+    input_hints: bool = Field(
+        default=False,
+        description="Whether to include input schema field information in system prompt",
     )
 
     # Token management
@@ -287,6 +291,27 @@ class InstructorBaseAgent[
         """
         self.memory.clear()
 
+    @property
+    def _input_schema_info(self) -> str:
+        """
+        Extract field names and descriptions from input schema.
+
+        Returns:
+            str: Formatted string with field information, empty if no input schema.
+        """
+        if not hasattr(self, "input_schema") or not self.input_schema:
+            return ""
+
+        field_info = []
+        for field_name, field_info_obj in self.input_schema.model_fields.items():
+            description = field_info_obj.description or "No description provided"
+            field_info.append(f"- **{field_name}**: {description}")
+
+        if not field_info:
+            return ""
+
+        return "\n".join(field_info)
+
     def _default_system_message(self) -> dict[str, str]:
         """
         Returns the default system message.
@@ -294,14 +319,21 @@ class InstructorBaseAgent[
         Returns:
             dict[str, str]: System message dictionary with role and content.
         """
+        content = self.system_prompt
+
+        # Add input schema hints if enabled
+        if self.input_hints:
+            input_info = self._input_schema_info
+            if input_info:
+                content += f"\n\nINPUT FIELD DESCRIPTIONS:\n{input_info}"
+
         return {
             "role": "system",
-            "content": self.system_prompt,
+            "content": content,
         }
 
     def _create_instructor_compatible_model(self, response_model: type[OutputSchema]):
         """Create a model that's compatible with instructor but avoids IOSchema validation."""
-        from pydantic import create_model
 
         # Get the fields from the original model
         fields = {}
--- a/akd/agents/_base.py
+++ b/akd/agents/_base.py
@@ -16,6 +16,7 @@ from pydantic import AnyUrl, BaseModel, Field, create_model, model_validator
 from akd._base import AbstractBase, BaseConfig, InputSchema, OutputSchema
 from akd.configs.project import CONFIG
 from akd.configs.prompts import DEFAULT_SYSTEM_PROMPT
+from akd.utils import get_model_fields
 
 
 class BaseAgentConfig(BaseConfig):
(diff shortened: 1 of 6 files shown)


akd/agents/_base.py
```diff
@@ -35,6 +36,10 @@ class BaseAgentConfig(BaseConfig):
         default=True,
         description="Whether to maintain conversation history/state",
     )
+    input_hints: bool = Field(
+        default=False,
+        description="Whether to include input schema field information in system prompt",
+    )
```
muthukumaranR, line 42: Not a review but dumping some thoughts as I think on this approach: 
I was thinking of these cases where this could be tricky:

1. the docstring having descriptive input / output hints that adds more semantics along with the descriptions - They could be more "tied-in" and integrated as a desctiptive hint rather than going by just field hints. problem is that then we need a protocol approach for adding in desriptions so the config can be universally applied. it may add more nuance to the developer experience too - now they have to think about one more thing in the framework. any idea on how we can automate this moving forward? I feel icky about using LLM for this either. 

2. would there be cases where you just want descriptions sometimes - or will IO hints always help? (for example - for the same config - we want to show the capability of the agents to the dashboard without adding in the IO hints - which are not useful to the end user / UIUX)

3. I am unsure of the behaviour of skip_no_description = True: Why would we skip them if there are no descriptions when the dev sets the hints flag to True? Should we not atleast provide the types atleast?

muthukumaranR (approved):
Added a discussion comment but nothing stopping this PR to be merged. thanks for this useful feature

akd/agents/_base.py
NISH1001: I agree.

1. I think LLM will complicate things. We could get description of the agent itself from the docstring of the Input Schema (the output schema is already being used and passed to LLM direclty as response model).
2. I think you're conflating the description of the agent with the system prompt enhancement. We can already get the description of the agent via its docstring no? `agent.__doc__` already gives a docstring for the agent to consume by any downstream like backend.  I guess a convinent way to do this is have an optional `description` field in the `BaseConfig` that we can override for any agent. That way, we could directly get `agent.description` as attribute. But adding the protocol although adds more control, but might complicate -- like you said -- for devs. That's why: originally I wanted the Field description to be used.
3. I think you're correct on that. I wasn't sure how to handle the situation as hinting in the prompt when there are no descriptions. This also means: there could be some params in the InputSchema that are completely irrelevant to LLM for generatoin but are used for internal logic. I am not sure how to handle this. I was thinking: instead of having a boolean, maybe we can have a list of field as hints and there, we can override which fields are important for LLM to "see". `input_hints: list[str] | None = .....` and if this is not empty, we can use those fields. What do you think?

muthukumaranR: on 2:

 I understand we can just get `__doc__` but I was thinking along the lines of config driven approach - rather than accessing the `__doc__`. I do agree `agent.description` is good approach - with the other setting will either add or ignore the IO hints to the `agent.description` based on the config. Some prompt string to consider the docstring IO docs when available would be a simple middle ground maybe? something like "prioritize the information in docstring if IO information is present" - or something more elegant than that

muthukumaranR: this way -  all the devs need to know is that we have Agent.desctiption available. alternatively - we can also add Agent.description_with_io_hints and pre-fill this field this irrespective of the config - that way, it's one less config to set and is always available when needed and used based on the discretion of the developer.

NISH1001: Update for (1):
- I have added `description` field in the base config. Previoulsy it was only availabe for tools. Now, it's available at `BaseConfig` level. Default behaviour when doing `agent.description` is now either from config else from docstring.

NISH1001: Update:
- The descritpion is also now part of the system prompt to enahnce if hinting is available.

NISH1001: I like the `description_with_io_hints` property. Maybe i can add this later. Thanks.

PR #232, opened 2025-10-01
Bugfix/code search tool

pranath-reddy:
### Summary :memo:
This PR fixes multiple issues in the Code Search Tool related to query storing, SDE search results, score normalization, and reranking logic. These changes improve result consistency when processing multiple queries.

### Details
- Fixed handling of the `query` field by explicitly attaching the query string to each result.  
- Corrected SDE search output by ensuring results are properly extended with associated queries.  
- Normalized CrossEncoder similarity scores using sigmoid function for stable values and Interpretability.  
- Updated reranking logic to rerank results per query and distribute `top_k` appropriately across multiple queries. 

### Bugfixes :bug: 
- Fixed missing or inconsistent `query` field in results.  
- Fixed SDE search output handling for multi-query scenarios.  
- Fixed unbounded CrossEncoder scores by normalizing them.  
- Fixed reranking to handle multi-query results correctly. 

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval

### DIFF
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -230,6 +230,7 @@ class CombinedCodeSearchTool(CodeSearchTool):
 
             # Get similarity scores from CrossEncoder
             scores = self.reranker_model.predict(pairs)
+            scores = 1 / (1 + np.exp(-scores))
 
             # Attach scores
             for score, result in zip(scores, results):
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -745,7 +745,6 @@ class SDECodeSearchTool(CodeSearchTool):
                             f"Error during search for query '{query}' on page {page}: {e}"
                         )
                         continue  # continue to the next page
-                all_results_data = all_results_data[: params.max_results]
             except Exception as e:
                 logger.error(f"Error during search for query '{query}': {e}")
 
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -734,6 +734,7 @@ class SDECodeSearchTool(CodeSearchTool):
                 for page in range(self.max_pages):
                     try:
                         results = self.sde_search(page=page, query=query)
+                        results = results[: params.top_k]
                         if results:
                             for result in results:
                                 result["query"] = query
--- a/akd/tools/code_search.py
+++ b/akd/tools/code_search.py
@@ -511,6 +511,8 @@ class LocalRepoCodeSearchTool(CodeSearchTool):
                     remove_embedding_column=self.config.remove_embedding_column,
                 )
                 if results:
+                    for result in results:
+                        result["query"] = query
                     all_results_data.extend(results)
             except Exception as e:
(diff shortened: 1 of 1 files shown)


akd/tools/code_search.py
```diff
@@ -214,8 +214,15 @@ async def _arun(
...
 
-        reranked = self._rerank_results(all_results, params.queries[0])[: params.top_k]
-        return self.output_schema(results=reranked, category="technology")
+        final_results = []
+        for query in params.queries:
+            query_results = [result for result in all_results if result.query == query]
+            reranked = self._rerank_results(query_results, query)[
+                : (params.top_k) // len(params.queries)
+            ]
+            final_results.extend(reranked)
```
NISH1001: Can this be made concurent? Normally asyncio handles it if we `gather(*tasks)`. I think it should be concurrrent as long as it doesn't impose any memory issue during reranking (which generates vector i guess). Does having concurrency impose memory issue if search results is too many?

NISH1001: Or should we concate the queries and rerank at just one time? I am not sure. You mentioned this multiple reranking leading to worser performance?

akd/tools/code_search.py
```diff
@@ -230,6 +237,7 @@ def _rerank_results(
 
             # Get similarity scores from CrossEncoder
             scores = self.reranker_model.predict(pairs)
+            scores = 1 / (1 + np.exp(-scores))
```
NISH1001, line 251: Oh interesting. Does this help?

akd/tools/code_search.py
```diff
@@ -511,6 +519,8 @@ async def _arun(
                     remove_embedding_column=self.config.remove_embedding_column,
                 )
                 if results:
+                    for result in results:
+                        result["query"] = query
                     all_results_data.extend(results)
```
NISH1001, line 535: Do we need if condition?

akd/tools/code_search.py
pranath-reddy: I'll make it concurrent, but at the tool level we shouldn't rerank all queries together, right? We might use the tool with a list of independent queries as well, not just the reformulated ones. We can rerank them all togther at agent level.

akd/tools/code_search.py
pranath-reddy: Just a safety check

akd/tools/code_search.py
pranath-reddy: No, the relative order remains same, but `predict()` was returning raw logits so just wanted to make the scores more readable.

akd/tools/code_search.py
NISH1001: We can make concurrent

pranath-reddy: Done

NISH1001:
@pranath-reddy 2 tests are failing. Can you retry the tests again? The `test_end_to_end_workflow_with_report_output` should pass I guess as it's intermittent.

NISH1001 (approved):

PR #235, opened 2025-10-01
Update Local Repo Search with Reformulated READMEs

pranath-reddy:
### Summary :memo:
This PR updates the repository dump used in the `LocalRepoCodeSearchTool` by integrating reformulated README content.  The reformulation was performed using the `ReadmeReformulationAgent`, which optimizes README text for vector search. Reformulated READMEs remove markdown artifacts, expand abbreviations, and enrich semantic content for improved retrieval relevance.

### Details
- Switched repository embeddings file from `repositories_with_embeddings_v4.csv` to `repositories_with_embeddings_v5.csv`.
- Updated dataset schema to include:
  - `reformulated_text` column for improved vector search context.  
  - `key_topics` column extracted during reformulation.
- Adjusted embedding generation to concatenate reformulated text, description, and key topics for richer search vectors.
- Refreshed Google Drive dataset link with latest dump.

Note: this PR does **not** add the reformulation agent itself, only updates the underlying dataset and code to leverage the reformulated data.

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval


akd/tools/code_search.py
```diff
@@ -251,19 +251,19 @@ class LocalRepoCodeSearchToolConfig(CodeSearchToolConfig):
...
     )
     embedder_type: Literal["sentence-transformers", "openai"] = "sentence-transformers"
     wait_time: int = 1
     embedding_model_name: str = os.getenv("CODE_SEARCH_MODEL", "thenlper/gte-large")
     remove_embedding_column: bool = True
-    text_column: str = "text"
-    name_column: str = "name"
+    text_column: str = "reformulated_text"
     desc_column: str = "description"
+    key_topics_column: str = "key_topics"
```
NISH1001: I was thinking: If this is a recurring thing for adding more and more columns: maybe what we can do is have list of columns as the config ? and concat accordingly? Instead of us adding new columns every now and then?

```python
columns: list[str] = Field(default_factory=lambda x: ["text"], ...)
```

and in the concat we can have
```python
def stringify_columns(self, columns: list[str]) -> str:
...
...
```

This might be robust and we don't need to create more and more fields in future?

akd/tools/code_search.py
```diff
@@ -381,6 +379,36 @@ def _ensure_data_file_exists(self):
...
             logger.info(f"Data file already exists at '{data_file_path}'.")
 
+    def _stringify_columns(self, columns: list[str]) -> list[str]:
+        """
+        Concatenate the given columns row-wise and return a list of embedding texts
+        """
+        # Defensive: check columns exist
+        missing = [c for c in columns if c not in self.repo_data.columns]
+        if missing:
+            raise KeyError(f"Missing columns in repo_data: {missing}")
```
NISH1001: Should we do this at initiliazioant level? Like after constructor within `_post_init()`?

That way we will know at compile time if the list is empty?

```python
def _post_init(...):
super()._post_init()
self._validate_columns(...)
```

NISH1001 (changes requested):

NISH1001 (approved):

PR #247, opened 2025-10-07
Bugfix scraped metadata

NISH1001:
Previously, in cases where some of the fields that weren't in ScrapedMetadata it'd throw error about those fields. Now, we segregate available fields and rest is added to extra


muthukumaranR (approved):
lgtm, thanks

PR #248, opened 2025-10-08
Reranker Tool [WIP]

pranath-reddy:
### Summary :memo:
This PR adds initial implementation of reranking tools for search results.

- A reranker tool base class for processing and reordering search results based on relevance scores.
- A CrossEncoder-based reranker tool that uses transformer models to compute query-document similarity for improved ranking.

### `RerankerTool` (Base Reranker Architecture)

- Introduced `RerankerTool` under `akd.tools.reranker` to provide a standardized interface for reranking search results.
- Uses `RerankerToolInputSchema` with a query string and list of `SearchResultItem` objects.
- Provides utility methods for deduplicating results by URL and sorting results by score.
- Adheres to the AKD tool format with proper schema validation.

### `CrossEncoderRerankerTool` (Reranking via Cross-Encoder)

- Configures a `CrossEncoder` instance using models from the sentence-transformers library (default: `cross-encoder/ms-marco-MiniLM-L12-v2`).
- Computes relevance scores between the query and each result's content using the cross-encoder model.
- Applies sigmoid normalization to scores for consistent 0-1 range values.
- Automatically deduplicates and sorts results by computed relevance scores.
- Integrated into `CombinedCodeSearchTool` for multi-tool result aggregation and reranking.

## Usage
```python
from akd.tools.reranker import CrossEncoderRerankerTool, CrossEncoderRerankerToolConfig

# Configure the reranker
config = CrossEncoderRerankerToolConfig(
    model_name="cross-encoder/ms-marco-MiniLM-L12-v2"
)

# Initialize the reranker tool
reranker = CrossEncoderRerankerTool(config=config, debug=True)

# Rerank search results
reranked_results = reranker._rerank_results(
    query="machine learning algorithms",
    results=search_results
)
```

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval


akd/tools/reranker.py
```diff
@@ -0,0 +1,144 @@
...
+from loguru import logger
+from sentence_transformers import CrossEncoder
+import numpy as np
+
+
+class RerankerToolConfig(BaseToolConfig):
+    """
+    Base configuration for reranker tools.
+    This can be extended by specific reranker tool configurations.
+    """
```
NISH1001, line 19: maybe add `model_name` as base tool level fieldname here since they can be applied for any model

akd/tools/reranker.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class RerankerToolInputSchema(InputSchema):
+    """
+    Schema for input to a tool for reranking search results.
+    """
+
+    query: str = Field(..., description="Reranking query.")
+    results: List[SearchResultItem] = Field(..., description="List of search results to rerank.")
```
NISH1001: use newer python 3.12 type hint style like `list[SearchResultItem]`

akd/tools/reranker.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class RerankerToolOutputSchema(OutputSchema):
+    """Schema for output of a tool for reranking search results."""
+
+    query: str = Field(..., description="Reranking query.")
+    results: List[SearchResultItem] = Field(..., description="List of search results to rerank.")
```
NISH1001: same as above

akd/tools/reranker.py
NISH1001, line 19: also maybe add some other params like `deduplication` boolean and defaults to True etc...

akd/tools/reranker.py
```diff
@@ -0,0 +1,144 @@
...
+        except TypeError:
+            # If sorting fails (mixed types), return as is
+            return results
+
+    # abstract method to be implemented by the subclass
+    def _rerank_results(self, query: str, results: List[SearchResultItem]) -> List[SearchResultItem]:
+        raise NotImplementedError("Subclass must implement this method")
+
+    def _arun(self, params: RerankerToolInputSchema) -> RerankerToolOutputSchema:
+        return self._rerank_results(params.query, params.results)
```
NISH1001: maybe what we can do is

```python
def _arun(...)
...
...
ranked_results = self._rerank....(...)
if self.deduplation:
ranked_results = self.dedecupliate(ranked_reuslts)

return ranked_results...
```

This way the base class itself has consistency...and we just have to implrvement raranking method itself....

NISH1001 (changes requested):

akd/tools/reranker.py
```diff
@@ -18,21 +16,26 @@ class RerankerToolConfig(BaseToolConfig):
     This can be extended by specific reranker tool configurations.
     """
 
+    use_deduplication: bool = Field(default=True, description="Whether to use deduplication of results.")
```
NISH1001: Let's rename this to deduplication? I like to think it as a parameter, so noun makes sense for me.

NISH1001 (changes requested):

akd/tools/reranker.py
```diff
@@ -0,0 +1,138 @@
...
+
+    query: str = Field(..., description="Reranking query.")
+    results: list[SearchResultItem] = Field(..., description="List of search results to rerank.")
+
+
+class RerankerToolOutputSchema(OutputSchema):
+    """Schema for output of a tool for reranking search results."""
+
+    query: str = Field(..., description="Reranking query.")
+    results: list[SearchResultItem] = Field(..., description="List of search results to rerank.")
```
NISH1001: I guess, the description is different for these? Eg: results are actually reranked result not "to rerank". Maybe copy-paste artifact haha

akd/tools/reranker.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    input_schema = RerankerToolInputSchema
+    output_schema = RerankerToolOutputSchema
+    config_schema = RerankerToolConfig
+
+    def _deduplicate_results(
+        self,
+        results: list[SearchResultItem],
+        key: str = "url",
```
NISH1001: maybe we can put `key` as part of config like `dedup_key` or something? So that we can control which key to use to deduplicate . Sometimes what we could do is not just url-based, but maybe title-based? Also wondering if we can do cascaded deduplication. Like the key in the config will look like list of keys defaulting to just `["url"]` but in cases where we know title might work, we could do `["url", "title"]` and this way we can do

```python
...
for _dkey in self.dedup_keys:
  results = self.deduplicate_results(results, key=_dkey)
```

akd/tools/reranker.py
```diff
@@ -0,0 +1,138 @@
...
+            val = str(getattr(result, key, ""))
+            if val and val not in seen:
+                seen.add(val)
+                deduped.append(result)
+        return deduped
+
+    def _sort_results(
+        self,
+        results: list[SearchResultItem],
+        sort_by: str = "score",
```
NISH1001: Same for `sort_by`. THink we can have more flexilibty to have this from config as well. SOmething like `sorting_key`?

NISH1001 (changes requested):

akd/tools/reranker.py
NISH1001: The reason is: we'd then only ahve to implement `_rerank...` method and the flow within _arun(...) won't have to change. And `_rerank` would be abstract method and has to be implemented downstream.

NISH1001:
```python

    def _arun(self, query: str, results: list[SearchResultItem]) -> list[SearchResultItem]:
        """
        Runs the reranking and deduplication flow.
        """
        # 1. Perform the core reranking logic
        ranked_results = self._rerank(query, results)

        # 2. Optionally deduplicate the results
        if self.deduplication:
            ranked_results = self._deduplicate(ranked_results)

        return ranked_results

    @abstractmethod
    def _rerank(self, query: str, results: list[SearchResultItem]) -> list[SearchResultItem]:
        """
        Abstract method for the specific reranking logic.
        This must be implemented by any subclass.
        """
        raise NotImplementedError()
```

akd/tools/code_search.py
```diff
@@ -208,7 +209,9 @@ async def _rerank_query(
     ) -> list[SearchResultItem]:
         """Rerank results for a single query."""
         query_results = [result for result in all_results if result.query == query]
-        return self._rerank_results(query_results, query)[:top_k_per_query]
+        reranked_results = self.reranker_tool._arun(RerankerToolInputSchema(query=query, results=query_results))
```
NISH1001: wait why are yo ucalling `_arun` here. It should be `arun` no?

NISH1001 (changes requested):

akd/tools/code_search.py
pranath-reddy: oh we use _arun right?

akd/tools/reranker.py
```diff
@@ -0,0 +1,150 @@
...
+
+class CrossEncoderRerankerTool(RerankerTool):
+    """
+    Tool for performing reranking of search results using a cross-encoder model.
+    """
+
+    def __init__(self, config: RerankerToolConfig | None = None, debug: bool = False):
+        super().__init__(config=config, debug=debug)
+        self.reranker_model = CrossEncoder(self.config.model_name)
+        self.debug = debug
```
NISH1001, line 140: Alternate way to make this minimal would be to override `_post_init()` I guess

```python

def _post_init(self) -> None:
  super()._post_init(...)
  self.reranker_model = CrossEncoder(self.config.model_name)
```

Just a FYI. No need to change.

NISH1001 (commented):
FYI

akd/tools/code_search.py
NISH1001: I mean internally yes. But at tool level, invocation is just `arun`. `_arun` is never intended to use at public scope. It's internal. Otherwise there will be no validation etc.. here it will be


```
reranker_tool -> arun -> |  implements _arun -> dowsntream flow -> | return output
```

The scope inside `|` here is meant as private scope which external user doesn't have to know about when using the tool. Calling _arun publicly will ignore all the validation for schema, etc...

pranath-reddy: oh okay got it

akd/tools/reranker.py
```diff
@@ -111,7 +111,7 @@ def __get_sort_key(result):
     def _rerank_results(self, query: str, results: list[SearchResultItem]) -> list[SearchResultItem]:
```
NISH1001: can we make reranking `async` as well.

NISH1001 (changes requested):

akd/tools/reranker.py
```diff
@@ -0,0 +1,150 @@
...
+    Attributes:
+        input_schema (RerankerToolInputSchema): The schema for the input data.
+        output_schema (RerankerToolOutputSchema): The schema for the output data.
+    """
+
+    input_schema = RerankerToolInputSchema
+    output_schema = RerankerToolOutputSchema
+    config_schema = RerankerToolConfig
+
+    def _deduplicate_results(
```
NISH1001: make async

akd/tools/reranker.py
```diff
@@ -0,0 +1,150 @@
...
+        for result in results:
+            for key in deduplication_keys:
+                val = str(getattr(result, key, ""))
+                if val and val not in seen:
+                    seen.add(val)
+                    deduped.append(result)
+                    break
+        return deduped
+
+    def _sort_results(
```
NISH1001: make async

akd/tools/reranker.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    Tool for performing reranking of search results using a cross-encoder model.
+    """
+
+    def __init__(self, config: RerankerToolConfig | None = None, debug: bool = False):
+        super().__init__(config=config, debug=debug)
+        self.reranker_model = CrossEncoder(self.config.model_name)
+        self.debug = debug
+
+    def _rerank_results(self, query: str, results: list[SearchResultItem]) -> list[SearchResultItem]:
```
NISH1001: async def

NISH1001 (changes requested):

NISH1001 (approved):

PR #251, opened 2025-10-09
Add Serper Search Tool and Enhancements

NISH1001:
**Title:** `feat: Add Serper Search Tool and Enhancements`

-----

### Summary :memo:

This pull request introduces the `SerperSearchTool`, a new capability for performing Google searches using the Serper API. This integration allows agents to query various search types, including general web, news, and images. The implementation includes optimizations to minimize API credit usage and enriches the search results with parsed dates, PDF links, and other relevant metadata.

### Major Changes :rocket:

  - **Added `SerperSearchTool`**: Integrated the Serper Google Search API as a new, fully-featured tool in `akd/tools/search/serper.py`. This includes the core tool class, Pydantic-based configuration, and input/output schemas.
  - **Optimized API Requests**: Refactored the tool's logic to make API requests more efficient, which significantly reduces the number of credits consumed per query.
  - **Improved Reliability**: Added a `pre_authenticate` flag and connection validation check during tool initialization to immediately confirm that the provided Serper API key is valid.

### Minor Changes :wrench:

  - **Added `extra` Field**: The base `SearchToolOutputSchema` in `akd/tools/search/_base.py` has been updated with an `extra` field to allow tools to return additional metadata.
  - **New `parse_date` Utility**: A `parse_date` helper function was added to `akd/utils.py` (using the new `dateparser` dependency) to standardize date strings from various formats into datetime objects.
  - **Enriched Serper Results**: The `SerperSearchTool` now uses the `parse_date` utility to populate the `published_date` and also extracts and populates a `pdf_url` if one is available in the search results.

### Usage :gear:

Here is an example of how to initialize and use the `SerperSearchTool`:

```python
    SerperSearchTool,
    SerperSearchToolConfig,
    SerperSearchToolInputSchema,
)

# Configure the tool
_cfg = SerperSearchToolConfig(
    # api_key="YOUR_API_KEY",  # Loaded from environment variables if not provided
    # pre_authenticate=True, # Optional: validates the API key on initialization
    debug=True,
)
search_tool = SerperSearchTool(_cfg)

# Run the search asynchronously
results = await search_tool.arun(
    SerperSearchToolInputSchema(
        queries=["landslide nepal"],
        category="science",
        max_results=20,
    )
)

# Print the results
print(results.model_dump())
```

### Checks

  - [x] Tested Changes (81% coverage on local)
  - [ ] Stakeholder Approval

### DIFF
--- a/akd/tools/search/serper.py
+++ b/akd/tools/search/serper.py
@@ -311,6 +311,7 @@ class SerperSearchTool(SearchTool):
                 result["query"] = query
                 if category:
                     result["category"] = category
+                result["page"] = page
 
             # Extract additional metadata fields for the extra field
             # Include everything except the results themselves
@@ -429,7 +430,7 @@ class SerperSearchTool(SearchTool):
             Tuple of (list of search result dictionaries, aggregated metadata dict)
         """
         all_results = []
-        aggregated_metadata = {}
+        all_page_metadata = []  # Collect metadata from each page
         current_page = 1
 
         while len(all_results) < target_results and current_page <= self.max_pages:
@@ -456,13 +457,9 @@ class SerperSearchTool(SearchTool):
                 all_results.extend(results)
                 all_results = await self._process_results(all_results)
 
-                # Merge metadata: keep first page's metadata for most fields
-                if current_page == 1:
-                    aggregated_metadata = metadata
-                else:
-                    # For list-based fields, we could append (but typically only page 1 has these)
-                    # For now, keep the first page's metadata as it's most relevant
-                    pass
+                # Collect metadata from each page for proper merging
+                if metadata:
+                    all_page_metadata.append(metadata)
 
                 current_page += 1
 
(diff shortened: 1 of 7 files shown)


akd/tools/search/serper.py
```diff
@@ -0,0 +1,630 @@
...
+
+    api_key: SecretStr = Field(
+        default_factory=lambda: SecretStr(os.getenv("SERPER_API_KEY", "")),
+        description="Serper API key for authentication",
+    )
+    base_url: HttpUrl = Field(
+        default=os.getenv("SERPER_BASE_URL", "https://google.serper.dev"),
+        description="Base URL for Serper API (without endpoint path)",
+    )
+    category: Literal["search", "scholar", "news", "images", "places"] = Field(
```
muthukumaranR, line 50: do we need all these categories?

akd/tools/search/serper.py
```diff
@@ -0,0 +1,630 @@
...
+    hl: str = Field(
+        default="en",
+        description="Language code for results (e.g., 'en', 'es')",
+    )
+    autocorrect: bool = Field(
+        default=True,
+        description="Whether to enable query autocorrection",
+    )
+    max_pages: int = Field(
+        default=int(os.getenv("SERPER_MAX_PAGES", "5")),
```
muthukumaranR: should this be 1 so we dont incur 5x credits by default?

akd/tools/search/serper.py
```diff
@@ -0,0 +1,630 @@
...
+        gt=0,
+        le=10,
+        description="Maximum number of pages to fetch per query",
+    )
+    result_multiplier: float = Field(
+        default=float(os.getenv("SERPER_RESULT_MULTIPLIER", "1.0")),
+        gt=0.0,
+        le=3.0,
+        description="Multiplier for over-fetching results to compensate for filtering (e.g., 1.5 fetches 50% extra). Use 1.0 to fetch exactly what's requested and save credits.",
```
muthukumaranR, line 92: haha, I have exactly the same parameter for searxNG after I added vllm reranking. I was assuming the good result is somewhere there and I just needed to dig deeper in the results but unfortunately that's not the case.

akd/tools/search/serper.py
```diff
@@ -0,0 +1,630 @@
...
+
+    input_schema = SerperSearchToolInputSchema
+    output_schema = SerperSearchToolOutputSchema
+    config_schema = SerperSearchToolConfig
+
+    _category_map = {
+        "science": "scholar",
+        "general": "search",
+        "technology": "search",
+    }
```
muthukumaranR, line 126: same as above - do we need this? claude thinks we need this?

akd/tools/search/serper.py
```diff
@@ -0,0 +1,630 @@
...
+        - Presence of knowledge graph data
+        - Result metadata quality
+
+        Args:
+            result: The result dictionary from Serper
+            position: Position in the result list (0-indexed)
+
+        Returns:
+            Normalized score between 0 and 1
+        """
```
muthukumaranR, line 349: not sure about this scoring strategy. base_score is straightforward but not sure about other things here

muthukumaranR: I'd just trust the google ranking process here if it were upto me.

akd/utils.py
```diff
@@ -259,3 +259,43 @@ def get_model_fields(
...
+    if date_input is None:
+        return None
+
+    # Handle year integers
+    if isinstance(date_input, int):
+        return datetime(date_input, 1, 1)
+
+    # Handle string dates
+    if isinstance(date_input, str):
+        parsed = dateparser.parse(date_input)
```
muthukumaranR: ah didnt know `dateparser` is a thing. nice.

muthukumaranR: would have saved more than a few LLM calls for a simple task across a few projects

akd/tools/search/serper.py
NISH1001: Good point.

akd/utils.py
NISH1001: Ya.

akd/tools/search/serper.py
NISH1001: Yes. It's meant as a placeholder since we already get the results in ranked order.

akd/tools/search/serper.py
NISH1001: I think at akd-level we'd only need `general` and `science`, hence mapped to search and scholar endpoint here.

akd/tools/search/serper.py
NISH1001: The downstream reranker would be able to assign proper score. This method here is meant as a placeholder and naive score mechanism, won't be useful but exist to maintain score float semantics. I didn't want to put None value so...

akd/tools/search/serper.py
NISH1001: It's to maintain serper original semantics. But I think we can just restrict to general and science.

muthukumaranR (approved):

PR #252, opened 2025-10-14
De-prioritize docling scraper in composite scraper to be last

NISH1001:
- `Crawl4aiScraper` has the highest priority
- DoclingScraper has the lowest priority


muthukumaranR (approved):
Agree with this order. Lgtm

PR #257, opened 2025-10-15
Bugfix: Add PDF Content Check to SimpleWebScraper

NISH1001:
**Description:**

### Summary :memo:
This pull request enhances the `SimpleWebScraper` to proactively detect when a URL points to a PDF file. Previously, the scraper would attempt to parse PDF content as HTML, leading to errors or incorrect output. Now, it raises a specific `RuntimeError` with a helpful message guiding the user to use a PDF scraper instead. This fixes the bug introduced by #252 

### Details
1.  **Content-Type Header Check**: In the `_fetch_webpage` method, the response's `Content-Type` header is now checked. If it contains `application/pdf`, the scraper immediately raises a `RuntimeError`.
2.  **Magic Byte Check**: As a fallback, the `_arun` method now inspects the first few bytes of the downloaded content. If the content begins with `%PDF-` (the magic number for PDF files), it also raises a `RuntimeError`. This handles cases where the server might not provide the correct `Content-Type` header.
3.  **Updated Docstrings**: The docstring for `_fetch_webpage` has been updated to include the new `RuntimeError` that can be raised.

### Bugfixes :bug: (delete if didn't have any)
- Fixes the issue where `SimpleWebScraper` would fail when provided with a URL that returns PDF content.

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval

### DIFF
--- a/akd/tools/scrapers/web_scrapers.py
+++ b/akd/tools/scrapers/web_scrapers.py
@@ -37,6 +37,7 @@ class SimpleWebScraper(WebScraper):
         Raises:
             HTTPError: If the HTTP request fails
             ValueError: If content length exceeds maximum
+            RuntimeError: If the content is a PDF
             RequestException: For other request-related errors
         """
         try:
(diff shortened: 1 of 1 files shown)


AwaleSajil (approved):
Approved LGTM

PR #269, opened 2025-11-04
feat: Add standardized response field to base OutputSchema

NISH1001:
### Summary :memo:

This PR introduces a standardized **`_response` computed field** to the base `OutputSchema`. This provides a consistent interface (e.g., `result._response`) to access the primary text output from any agent, regardless of the specific field name used in the subclass (like `report`, `output`, etc.).

This simplifies consuming agent results, as the calling code no longer needs to know the specific structure of each agent's output schema.

I have named `_response` because in most of the tests `response` was the standard field for llm extraction. I didn't wnat to have to resolve tons of those conflicts.

### Details

  * **`akd/_base.py`**: Added a `@computed_field` named `_response` to the `OutputSchema`. This field's value is dynamically read from the field name specified in a class's `__response_field__` attribute. By default, all the downstream outputschema (if not set) will have empty string for the `_response`.
  * **`akd/agents/search/_base.py`**: Updated `SearchAgentOutputSchema` to use this new feature by setting `__response_field__ = "report"`.
  * **`akd/agents/gap_analysis/gap_analysis.py`**: Updated `GapOutputSchema` to set `__response_field__ = "output"`.
  * **`tests/`**: Added new unit tests to `test_deep_search.py`, `test_gap_agent.py`, and `test_code_search_tool_test.py` to verify that the `_response` field correctly mirrors the content of the designated primary field and gracefully handles `None` values.

### Usage

To make any new agent's output compatible, simply set the `__response_field__` class attribute in its `OutputSchema` to point to the desired string field.

Example `HaikuAgent`

```python
from akd.agents._base import InputSchema, OutputSchema, LiteLLMInstructorBaseAgent
from pydantic import Field

class HaikuAgentInputSchema(InputSchema):
    """
    Input schema that takes in query
    """
    query: str = Field(
        ...,
        description="Query to the haiku agent.",
    )
    topic_steer: str = Field(
        default="5-3-5 format",
        description="Steer in 5-3-5 syllable as possible. Also use sanskrit words in the middle line",
    )

class HaikuAgentOutputSchema(OutputSchema):
    """
    Generate a haiku from given query
    """
    # This line points the _response field to the 'haiku' field
    __response_field__ = "haiku"

    haiku: str = Field(..., description="Haiku content")

class HaikuAgent(LiteLLMInstructorBaseAgent):
    input_schema = HaikuAgentInputSchema
    output_schema = HaikuAgentOutputSchema

# ---
# When the agent is run:
# agent = HaikuAgent()
# result = await agent.arun(...)
#
# # We can now consistently access the text output
# print(result._response) 
#
# # This is now equivalent to:
# print(result.haiku)
```

### Checks

  - [x] Tested Changes (New unit tests added)
  - [ ] Stakeholder Approval


muthukumaranR (approved):
thanks for the field update. looks good to me

PR #271, opened 2025-11-10
Add Cross-Encoder Reranking to Search Agent

pranath-reddy:
### Summary :memo:
This PR integrates reranking capabilities into the search agent architecture by adding cross-encoder reranking support at the base SearchAgent level. The reranker improves search result quality by reordering results based on semantic relevance to the query, and can be optionally disabled by setting the reranker type to "none".

### Details
- Added reranker configuration fields to `SearchAgentConfig`:
  - `reranker_type: RerankerType` with default value "none" to specify the type of reranker
  - `reranker_config: RerankerToolConfig` with default factory for cross-encoder configuration (model: "cross-encoder/ms-marco-MiniLM-L12-v2")
- Integrated reranker imports and tooling from `akd.tools.reranker` module:
  - `RerankerTool` for executing reranking operations
  - `RerankerToolConfig` for configuration
  - `RerankerType` enum for specifying reranker model
  - `create_reranker` factory function
  - `RerankerToolInputSchema` for input validation
- Added reranker initialization in `SearchAgent.__init__()` method, making it available to all search agent subclasses
- Updated `CodeSearchAgentConfig` to inherit reranker configuration with default type "cross_encoder"
- Implemented reranking logic in `ControlledSearchAgent._run()` method:
  - Reranks accumulated results after each search iteration
  - Uses `RerankerToolInputSchema` to structure reranker input with query and results
  - Replaces `all_results` with reranked results to improve result ordering for subsequent iterations

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval


akd/agents/search/code_search.py
```diff
@@ -66,6 +66,12 @@ class CodeSearchAgentConfig(ControlledSearchAgentConfig):
     followup_query_prompt: str = Field(default=CODE_QUERY_PROMPT, description="System prompt for follow-up query agent")
     relevancy_prompt: str = Field(default=CODE_RELEVANCY_PROMPT, description="System prompt for relevancy agent")
 
+    # Reranker configuration
+    search_result_reranking: bool = Field(
```
NISH1001: Maybe just doing reranking parameter? I think we don't even need this because if we have `reranker_type` already, and setting it to `none` or `nope`, it iwll use NoOp reranker (no changes). 

in create_reranker function...

```python
    if reranker_type in ("identity", "no_op", "nope", "none"):
        return NoOpRerankerTool(config=config, debug=debug)
```

akd/agents/search/controlled.py
```diff
@@ -120,6 +127,22 @@ class ControlledSearchAgentConfig(LitSearchAgentConfig):
         description="Relevancy score threshold to trigger full content fetching",
     )
 
+    # Reranker configuration
+    search_result_reranking: bool = Field(
+        default=False,
+        description="Enable reranking of search results.",
+    )
```
NISH1001: I don't think we need it. This can easily be done through `reranker_type` and setting to `none` or `nope` will do it.

create_reranker function
```python
    if reranker_type in ("identity", "no_op", "nope", "none"):
        return NoOpRerankerTool(config=config, debug=debug)

```

akd/agents/search/controlled.py
```diff
@@ -850,6 +879,14 @@ async def _arun(
 
             all_results.extend(current_results)
 
+            if self.config.search_result_reranking:
```
NISH1001: We could avoid this conditional check and direclty apply reranking. Because if it was `none` or `nope` type or osmething, doing `reranker.arun(...)` won't affect anything with original list.
```

reranker_input = RerankerToolInputSchema(...)
reranked_results = await self.reranker.arun(reranker_input)
all_results = reranked_results.results
```

This avoids conditional and uses state-machine-like pattern based on reranker_type.

NISH1001 (changes requested):

akd/agents/search/controlled.py
pranath-reddy: Ah yes. got it. btw why do we have `nope` in there XD

NISH1001: It was an easter egg. lmao. i like to put some hidden funny things in code bases.

NISH1001:
@pranath-reddy Should we add this at `SearchAgent` level at base class in `akd.agents.search._base.py` just thei nitialization etc....and we can add the logic in controlled search where to rerank? I think making it available at base search level, we can access it in any of the subclasses?

NISH1001 (approved):

PR #272, opened 2025-11-11
Improve type abstraction for search results by introducing SearchResult base data structure

NISH1001:
### Summary :memo:
This pull request introduces a new base class, **`akd.structures.SearchResult`**, to create a more general abstraction for search results. The existing `SearchResultItem` (which is specific to literature/web search) has been refactored to inherit from this new base class.

This change allows the base `SearchAgent` and its related components to operate on the more generic `list[SearchResult]` type. This makes the agents more flexible and easier to adapt for different types of search (e.g., code search, database search) in the future.

Additionally, the `iterations_performed` field has been removed from the main `SearchAgentOutputSchema` and moved into the `extra` dictionary to keep the primary schema cleaner.

### Details
1.  **`akd/structures.py`**:
    * Introduced the new `SearchResult` base class, which contains common fields like `query`, `title`, `content`, `score`, and `extra`.
    * Refactored `SearchResultItem` to inherit from `SearchResult` and removed the duplicated fields.
2.  **`akd/agents/search/_base.py` & `akd/agents/search/answer.py`**:
    * Updated type hints in `SearchAgentOutputSchema`, `_generate_answer`, `_generate_report`, and `QuestionAnsweringAgentInputSchema` to use `list[SearchResult]` instead of the more specific `list[SearchResultItem]`.
    * Removed the `iterations_performed` field from `SearchAgentOutputSchema`.
3.  **`akd/agents/search/controlled.py` & `akd/agents/search/deep_search.py`**:
    * Updated the instantiation of `SearchAgentOutputSchema` to move `iterations_performed` into the `extra` dictionary.
4.  **`tests/`**:
    * Updated `test_deep_search.py` and `code_search_tool_test.py` to reflect the removal of the `iterations_performed` field from the output schema.
    * Tests that check this value now access it from the `result.extra` dictionary (e.g., `result.extra.get("iterations_performed")`).

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval

### DIFF
--- a/akd/structures.py
+++ b/akd/structures.py
@@ -36,22 +36,54 @@ except ImportError:
 # =============================================================================
 
 
-class SearchResultItem(IOSchema):
+class SearchResult(IOSchema):
+    """Base class for all search result types across different search domains.
+
+    Provides common interface for search results including query context,
+    title, content/description, and relevance scoring. Subclasses should
+    add domain-specific fields (e.g., url, doi for literature; decompositions
+    for hierarchical data search).
+    """
+
+    query: str = Field(..., description="Query that produced this search result")
+    title: str = Field(..., description="Title or name of the search result")
+    content: str = Field(default="", description="Content snippet, description, or summary")
+    score: float | None = Field(None, description="Relevance or ranking score")
+    extra: dict[str, Any] = Field(default_factory=dict, description="Extra metadata")
+
+    @computed_field
+    def relevancy_score(self) -> float | None:
+        """Alias for score field for consistency."""
+        return self.score
+
+    @field_validator("content", mode="before")
+    @classmethod
+    def validate_content(cls, v):
+        """Convert None to empty string for content field."""
+        return v if v is not None else ""
+
+    @field_validator("score", mode="before")
+    @classmethod
+    def validate_score(cls, v):
+        """Convert numpy float types to Python float for JSON serialization."""
+        if v is None:
+            return None
+        # Handle numpy float types (float32, float64, etc.)
+        if hasattr(v, "item"):  # numpy scalar types have .item() method
+            return float(v.item())
+        return float(v)
+
+
+class SearchResultItem(SearchResult):
     """Represents a single search result item with metadata."""
 
     # Required fields
     url: AnyUrl = Field(..., description="The URL of the search result")
-    title: str = Field(..., description="The title of the search result")
-    query: str = Field(..., description="The query used to obtain the search result")
 
     # Optional metadata
     pdf_url: AnyUrl | None = Field(
         None,
         description="The PDF URL of the search paper",
-    )
-    content: str = Field(
-        default="",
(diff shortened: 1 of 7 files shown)


NISH1001:
I was hesitant on renaming `SearchResultItem` because of cascaded effect because this data structure is what lit and code search uses intensively. And inheriting data-search-related data structure from `SearchResultItem` doesn't make sense. So, instead: introducing a more general type base `SearchResult` solves it and we can have new set of data-related structure (eg: `DataSearchResult`)

NISH1001:
---

Potential downstreams

```mermaid
classDiagram
    %% Existing Design (Before)
    class SearchResultItem {
        +str query
        +str title
        +str content
        +float score
        +dict extra
    }

    class SearchAgentOutputSchema {
        +List~SearchResultItem~ results
        +int iterations_performed
    }

    SearchAgentOutputSchema --> SearchResultItem : uses


    %% Refactored Design (After)
    class SearchResult {
        +str query
        +str title
        +str content
        +float score
        +dict extra
    }

    class SearchResultItem {
        +source: str
        +url: str
    }

    class DataSearchResult {
        +data_id: str
        +metadata: dict
    }

    class SearchAgentOutputSchemaNew {
        +List~SearchResult~ results
        +dict extra
    }

    SearchResultItem --|> SearchResult : inherits
    DataSearchResult --|> SearchResult : inherits
    SearchAgentOutputSchemaNew --> SearchResult : uses
```

---

### 🧩 Explanation

* **Before:**

  * `SearchAgentOutputSchema` directly depended on `SearchResultItem`, which was specific to literature/web search.
  * Any new type of search (like code or data search) required rewriting or duplicating similar structures.

* **After:**

  * Introduced a **base class `SearchResult`** containing shared fields (`query`, `title`, `content`, `score`, `extra`).
  * Specialized result types (e.g., `SearchResultItem`, `DataSearchResult`) can **extend** it for domain-specific fields.
  * `SearchAgentOutputSchema` now depends on the **generic type `List[SearchResult]`**, making it **more extensible and maintainable**.

akd/structures.py
```diff
@@ -36,23 +36,55 @@
...
+    """
+
+    query: str = Field(..., description="Query that produced this search result")
+    title: str = Field(..., description="Title or name of the search result")
+    content: str = Field(default="", description="Content snippet, description, or summary")
+    score: float | None = Field(None, description="Relevance or ranking score")
+    extra: dict[str, Any] = Field(default_factory=dict, description="Extra metadata")
+
+    @computed_field
+    def relevancy_score(self) -> float | None:
```
muthukumaranR, line 55: I think ranking score and relevancy score need to be different by default.

muthukumaranR (approved):
Thanks, lgtm

akd/structures.py
NISH1001: Agreed. do we separate it out as optional field then? what do you tihnk?

PR #273, opened 2025-11-12
Add Query Division Classifier to Code Search

pranath-reddy:
### Summary :memo:
This PR adds query classification to the code search agent and tools by implementing an LLM-based division classifier that categorizes user queries into one of NASA's five science divisions. This enhancement allows the search agent to narrow down searches to repositories belonging to specific divisions, improving search precision and relevance.

### Details
- Created `ScienceDivision` enum and `DivisionAgent` class in `intents.py` for classifying queries into Earth Science, Planetary Science, Astrophysics, Heliophysics, Biological and Physical Sciences, or Unknown divisions
- Added comprehensive `DIVISION_PROMPT` in `code_prompts.py` with detailed overviews, study areas, and mission examples for each division
- Enhanced `LocalRepoCodeSearchTool` and `SDECodeSearchTool` to support division-based filtering with `use_division_filter_local` and `use_division_filter_sde` configuration flags in the agent
- Modified `find_repo()` and `sde_search()` methods to be async, perform division classification, and filter results by division
- Updated `CodeSearchAgentConfig` to inherit division filter configuration and automatically initialize division agent when enabled
- Added query division tracking in result metadata via `extra` field for transparency

### Usage
```python
from akd.agents.intents import DivisionAgent, DivisionInputSchema
from akd.agents._base import BaseAgentConfig
from akd.configs.code_prompts import DIVISION_PROMPT

# Configure and initialize the agent
cfg = BaseAgentConfig(model_name="gpt-4o-mini", system_prompt=DIVISION_PROMPT)
agent = DivisionAgent(config=cfg)

# Run the search
result = await agent.arun(
    DivisionInputSchema(
        query="landslide nepal", 
    )
)
```

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval


akd/tools/search/code_search.py
```diff
@@ -84,6 +93,12 @@ class CodeSearchTool(SearchTool):
     output_schema = CodeSearchToolOutputSchema
     config_schema = CodeSearchToolConfig
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.division_agent = DivisionAgent(
+            config=BaseAgentConfig(model_name="gpt-4o-mini", system_prompt=DIVISION_PROMPT)
+        )
```
NISH1001, line 100: I thought this would be part of the agent not tool.

Something like:
```python

agent = CodeSearchAgent(division_classifier=....,...)
```

And the filter would be applied at agentic level. At tool-level what we can do is, provide enum as a way to just filter without any classification.

```python

tool = CodeSearchTool(Config(division_filter=True,...))

await tool.arun(CodeInputSchema(query=..., division="Something")) # where we pass division from upstream. It could be None as well. If `division` is None, then no need to filter, else just filter using that.
```

NISH1001: I'd also try to avoid complicating the tools itself. If we have this as part of agent, few things would be nice:
- Division classification reasoning trace from original query
- We can use that trace  and division to drive subsequent agentic iteration

NISH1001: Let's have it as part of CodeSearchAgent, not tool. I'd avoid modifying tool to have llm-stuff unless necessary. This way, divsion classifier wil lbe an intent udnerstanding layer at agent layer,..,,

NISH1001 (changes requested):

akd/tools/search/code_search.py
NISH1001: Reiterating:

```python

division = division_classifier.arun(query....) if self.division_classifier
results_full = []
while iteration...:
  results = self.tool.arun(..., division=division.....) # if division is there, else Just None filter might work
# We need to test if performing classification at each iteration results in same set of division or not. Otherwise, original intent might drive the iteration.
# Also, we could reuse the same intent udnerstanding to udnerstand the results as well.
# Same division classifier -> run through each result -> get division -> then filter
# Doing this: we know tool is already dterministic, and the tuning is just happening at agent level.
relevant_results = self.checker.arun(results) # does relevancyc hecks tuned to code search itself. Something like archetype we have for lit and data search

results_full.extend(relevant_results)

refomrulation = self.reformulate(relevant_results, division_traces, division=....) # this will play nice with reformulation as well because there will be more context from the intent/division that can drive next iteration.
- perform next iteration or stop, etc

  
```

This way, tool behaviour is deterministic in nature and all the non-determinism comes from llm/agent workflow. And this will make a case taht agent improved because we did classification, then tool call, then filter, then checks, anything. Tool tentatively remains the same and only we focus on agentic improvement. This avoids pinpointing non-determinism. If tool is non-deterministic, then we'd have hard time replicating tool call results because there's llm layer to it. This is also the problem I have with `LinkAssessor` as well. Replication is hard with LinkAssessor (but there's leeway: because we're giving so many urls, 10s of urls, to it, it's fine if there's some non-determinism to it; just a side note).

This is also the reason we're not using Searchpipeline in akd lit v2 because searchpipeline was kind of non-deterministic. But again, since it's owrking with 100s of urls, it's less non-deterministic in nature stastiticaly.

pranath-reddy: Yeah, makes sense

NISH1001: But thank you for the work. You're on right track. Let's segregate non-dtermnisim more to agentci worfklow than tool. That way: tool call will be our baselines always and always replicable. The comments are feedbacks, not discouraging you to think about the process.

pranath-reddy: Yup, got it. I'll continue working on this PR after tomorrow's SME session.

PR #274, opened 2025-11-12
Add Per-Link Relevancy Assessment to Controlled Search

pranath-reddy:
### Summary :memo:
This PR integrates per-link relevancy assessment into the code search workflow, allowing fine-grained filtering of search results based on their relevance to the original query. The `LinkRelevancyAssessor`, which was previously imported but unused in `ControlledSearchAgent`, is now actively filtering results before they are added to the final result set.

**Note**: Current experiments show this feature **significantly degrades performance** (both accuracy and MRR). Additional experiments and tuning are required before enabling this by default. The feature is disabled by default.

### Details
- Added `enable_per_link_assessment` configuration flag to both `CodeSearchAgentConfig` and `ControlledSearchAgentConfig` (default: `False`)
- Initialized `LinkRelevancyAssessor` in `ControlledSearchAgent.__init__()` with debug configuration
- Added `LinkRelevancyAssessorInputSchema` import to support per-link assessment workflow
- Integrated link relevancy filtering in the search iteration loop: filters `current_results` using the assessor before adding to `all_results`
- Filters are applied based on the original query, with optional reformulated query and domain context support
- Added debug logging to track the number of results before and after link relevancy assessment

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval


akd/agents/search/controlled.py
```diff
@@ -156,6 +163,8 @@ def __init__(
         self.query_agent = query_agent or QueryAgent()
         self.followup_query_agent = followup_query_agent or FollowUpQueryAgent()
 
+        self.link_relevancy_assessor = LinkRelevancyAssessor(config=LinkRelevancyAssessorConfig(debug=self.debug))
```
NISH1001, line 166: We need to fine-tune the prompt for RelevancYAgent the LinkRelevancyAssessor has access to specifically for code search. Can we do that? LinkRelevancYAssessor does take in RelevancyAgent object as dependency as well.

IDeally, linkassessor should be tuned for code search here to work correctly.

NISH1001 (changes requested):

akd/agents/search/controlled.py
NISH1001: We can revisit this after SME discussion on how do we assess the content of repo.

PR #278, opened 2025-11-19
Feature/llm reranker

iamsims:
## Summary
- Implements LLM-based reranker that evaluates search results using configurable criteria, weights and fields. 
- In addition to sorting provides the label for each critera 
- Two implemnetations:
        -  Single LLM call per result evaluates all criteria simultaneously commit hash (02eb523)
        - Multiple LLM call per result (for per criteria)  commit hash (166f475)
- Supports weighted multi-criteria scoring with customizable categories

### DIFF
--- a/akd/tools/reranker.py
+++ b/akd/tools/reranker.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 from abc import abstractmethod
 from typing import Any, Literal
 
@@ -196,26 +195,16 @@ class ScoringCriterion(BaseModel):
     )
 
 
-class IndividualScore(OutputSchema):
-    """Score output from LLM for a single result against a single criterion."""
+class CriterionScore(BaseModel):
+    """Score for a single criterion."""
 
-    criterion: str = Field(..., description="The criterion being evaluated")
     category: str = Field(..., description="The selected category (e.g., 'Perfect Match', 'Acceptable', 'Unusable')")
     reasoning: str = Field(..., description="Brief explanation for the score")
-
-
-class IndividualScoreInput(InputSchema):
-    """Input schema for scoring a single result."""
-
-    query: str = Field(..., description="The search query")
-    result_content: dict[str, Any] = Field(..., description="The result fields to evaluate")
-    criterion: ScoringCriterion = Field(..., description="The criterion to evaluate against")
 
 
 class LLMRerankerToolConfig(RerankerToolConfig):
     """Configuration for LLM-based reranker."""
 
-    # LLM Agent configuration
     base_url: AnyUrl | None = Field(default=None, description="Base URL for LLM API")
     api_key: str | None = Field(default=None, description="API key for LLM")
     model_name: str = Field(default="gpt-4o-mini", description="LLM model name")
@@ -223,8 +212,8 @@ class LLMRerankerToolConfig(RerankerToolConfig):
     agent_system_prompt: str = Field(
         default=(
             "You are an expert at evaluating search results. "
-            "Analyze the provided result for the query against the given criterion and "
-            "select the most appropriate category. Provide clear reasoning."
+            "Analyze the provided result for the query against all given criteria and "
+            "select the most appropriate category for each. Provide clear reasoning."
         ),
         description="System prompt for the internal scoring agent",
     )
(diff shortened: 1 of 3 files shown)


NISH1001:
@iamsims add the detail `# Usage` sectoin as well in the pr description.

See some of the older PRs we have done:
eg #279

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+    model_name: str = Field(default="gpt-4o-mini", description="LLM model name")
+    temperature: float = Field(default=0.0, ge=0.0, le=2.0, description="LLM temperature")
+    agent_system_prompt: str = Field(
+        default=(
+            "You are an expert at evaluating search results. "
+            "Analyze the provided result for the query against all given criteria and "
+            "select the most appropriate category for each. Provide clear reasoning."
+        ),
+        description="System prompt for the internal scoring agent",
+    )
```
NISH1001, line 217: have this get from os.getenv as well so that we can globally modify whenever we want...and default to this value you have. something like `AKD_LLM_RERANKER_SYSTEM_PROMPT`

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+            ScoringCriterion(
+                name="Relevancy",
+                description="How relevant is this result to the query?",
+                weight=1.0,
+            ),
+            ScoringCriterion(
+                name="Processing Level",
+                description="How well does this result match the required processing level?",
+                weight=0.5,
+            ),
```
NISH1001, line 238: This seems too specific to data search agent. we need to find some general criteria that can be globally applied to any usecase...

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+    ):
+        """
+        Initialize the LLM reranker.
+
+        Args:
+            config: Reranker configuration
+            debug: Enable debug logging
+        """
+        super().__init__(config=config, debug=debug)
+        self.config: LLMRerankerToolConfig = self.config  # type hint
```
NISH1001, line 296: we don't need this...it will automatically be done through super()...

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+        )
+
+        class DummyInput(InputSchema):
+            """Dummy input schema for scoring agent."""
+
+            pass
+
+        class ScoringAgent(LiteLLMInstructorBaseAgent):
+            input_schema = DummyInput
+            output_schema = dynamic_scoring_model
```
NISH1001, line 321: move this to anotehr internal method like `_setup_scoring_agent` or something

and we can just do `self.scoring_agent = self._setup_scoring_agent(model....)

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+        return desc
+
+    def _create_dynamic_scoring_model(self, criteria: list[ScoringCriterion]) -> type[BaseModel]:
+        """
+        Dynamically create Pydantic model with one field per criterion.
+
+        Similar to relevancy-ranker.py approach - creates explicit named fields
+        for each criterion so LLM can see them in the JSON schema.
+        """
+        from pydantic import create_model
```
NISH1001, line 348: move to top-level import

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+        """
+        # Format the prompt with all criteria
+        formatted_prompt = self._format_prompt_template(
+            query=query,
+            result_content="\n".join(f"{k}: {v}" for k, v in result_content.items()),
+            criteria=self.config.scoring_criteria,
+        )
+
+        if self.debug:
+            print(formatted_prompt)
```
NISH1001, line 457: logger.debug(...)

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+        try:
+            messages = [
+                self.scoring_agent._default_system_message(),
+                {
+                    "role": "user",
+                    "content": formatted_prompt,
+                },
+            ]
+
+            print(messages)
```
NISH1001, line 468: remove print stattement...or add `if self.debug: logger.debug(...)`

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+
+    def __str__(self) -> str:
+        criteria_names = [c.name for c in self.config.scoring_criteria]
+        return f"{self.__class__.__name__} | criteria={criteria_names}"
+
+
 def create_reranker(
     reranker_type: RerankerType,
-    config: RerankerToolConfig | None = None,
+    config: RerankerToolConfig | LLMRerankerToolConfig | None = None,
```
NISH1001, line 578: Technically don't need `LLMRerankerToolConfig` because it's also a type of `RerankerToolConfig`. Redundant type hint

akd/tools/reranker.py
```diff
@@ -153,9 +156,428 @@ def __str__(self) -> str:
...
+            ]
+
+            print(messages)
+
+            response = await self.scoring_agent.get_response_async(
+                messages=messages,
+            )
+
+            results = {}
+            response_dict = response.model_dump()
```
NISH1001, line 475: what if we do this at `agent.arun` level? since it's one-level higher abstraction. is it possible to do it with the formatted_prompt? you have to convert it to ap ydantic input schema

NISH1001 (changes requested):
Initial review

NISH1001:
@iamsims let's visit this later. Low priority

PR #280, opened 2025-11-20
Improve Exposed Parameter Type Inference and Add Setter Validation

NISH1001:
### Summary

This PR significantly enhances the **`exposed_param`** decorator and the parameter introspection capabilities of the `ParamExposureMixin`.

The main improvements include:
1.  **More Robust Type Inference:** The `get_exposed_params` method now attempts to infer the parameter type from **`fget` (getter) return hints** first, and falls back to **`fset` (setter) parameter hints** before resorting to runtime value inspection. This ensures more accurate and reliable type information, especially for properties that might not have a value set yet.
2.  **Automatic Setter Type Validation:** A new `ValidatedProperty` class is introduced. The `@exposed_param` decorator now uses this class, which automatically wraps the property's setter with logic to **validate the input value against the property's type hints**, preventing accidental assignment of wrong types and raising a clear `TypeError`.
3.  **New Data Structure for Exposed Params:** The return type of `get_exposed_params` is changed from a raw `dict` to a `list[ExposedParamRuntimeInfo]`. This new `ExposedParamRuntimeInfo` Pydantic model provides a clean, validated, and structured way to return all metadata, including the new `type_source` field, which indicates *how* the type was inferred (`fget`, `fset`, or `runtime`).

### Changes Detailed

* **`akd/_base/utils.py`:**
    * Added `ExposedParamRuntimeInfo` Pydantic model for structured parameter metadata retrieval.
    * Introduced `ValidatedProperty` to automatically add type validation logic to property setters. This logic checks against `fget` return hints or `fset` parameter hints, handling both simple types and `Union` types.
    * Updated the `@exposed_param` decorator to use `ValidatedProperty` instead of the base `property`.
    * Refactored `ParamExposureMixin.get_exposed_params` to use the new type inference logic (prioritizing type hints) and return a list of `ExposedParamRuntimeInfo` objects.

* **`tests/agents/base/test_exposed_params.py`:**
    * Updated tests to expect and handle the new `list[ExposedParamRuntimeInfo]` return type for `get_exposed_params`.
    * Added a new test `test_exposed_param_type_validation` to confirm that the `ValidatedProperty` correctly raises a `TypeError` when an attempt is made to set a value with an incorrect type.
    * Updated existing tests to check for the new `type_` field instead of the old `type` key in the dictionary.

### Motivation

This change makes the **`exposed_param`** feature more reliable and developer-friendly by:

1.  **Improving discoverability:** Providing more accurate type information via type hints.
2.  **Enhancing stability:** Preventing runtime errors caused by passing values of the wrong type to setters.

---


akd/_base/utils.py
```diff
@@ -125,35 +125,70 @@ def get_exposed_params(self, include_values: bool = False) -> dict[str, dict[str
...
+
+        def __get_type_from_hints(attr: property) -> str:
+            """Extract type from property type hints. Priority: fget return > fset param."""
+            # Try fget return type
+            try:
+                hints = get_type_hints(attr.fget)
+                if "return" in hints:
+                    return getattr(hints["return"], "__name__", str(hints["return"]))
+            except Exception:
+                pass
```
muthukumaranR: can we log the error so it's easier to observe / fix if something breaks?

akd/_base/utils.py
```diff
@@ -125,35 +125,70 @@ def get_exposed_params(self, include_values: bool = False) -> dict[str, dict[str
...
                 - current_value: Current value (if include_values=True)
         """
+
+        def __get_type_from_hints(attr: property) -> str:
+            """Extract type from property type hints. Priority: fget return > fset param."""
+            # Try fget return type
+            try:
+                hints = get_type_hints(attr.fget)
+                if "return" in hints:
+                    return getattr(hints["return"], "__name__", str(hints["return"]))
```
muthukumaranR: `str` on types like Union / Optional may be problematic and include them too - we may want to resolve it into actual types

akd/_base/utils.py
```diff
@@ -125,35 +125,70 @@ def get_exposed_params(self, include_values: bool = False) -> dict[str, dict[str
...
             attr = getattr(type(self), attr_name, None)
-            if isinstance(attr, property) and hasattr(attr.fget, "_exposed_meta"):
-                meta = attr.fget._exposed_meta
+            if not (isinstance(attr, property) and hasattr(attr.fget, "_exposed_meta")):
+                continue
+
+            meta = attr.fget._exposed_meta
+
+            # Get type from hints, fallback to runtime value
+            param_type = __get_type_from_hints(attr)
```
muthukumaranR:
```suggestion
            param_type = _get_type_from_hints(attr)
```

muthukumaranR (approved):

akd/_base/utils.py
NISH1001: Ya. But i think we're not actually using the type information anywhere currently...

akd/_base/utils.py
NISH1001: done

PR #287, opened 2025-11-22
bugfix: Refactor config synchronization using Metaclass dynamic properties

### LINKED ISSUE #286: Nasty bug when we start exposing internal paramters to external systems
### Describe the bug

just found one nasty bug/side-effect in akd core..it's nota bug per se

right now, when we pass config....all the config values/items are also direclty asigned to object level attribute using setattr. That is: obj.config.x is also set at obj.x. But problem is: referencing problem with primitive types like int, str, bool, etc... (not list, set, etc...). So if x was int: and we did obj.x = <new_int>, it will not be refrlected to obj.config.x since int are not assigned by reference. 
This bug is not apparent if we are not really modifying any udnerlying config values at runtime... a large part of the code I have wrriten direclty uses obj.x  notation instead of obj.config.x when working on internal logic
But will start to surface if we start exposing some of these paramters. Right now, with @exposed_param we can manually point to config.x 

Again, this bug is only apparent when we start exposing internal paramters to external system. My mental model initially was: everything will be config-driven. So during object-creation time, we will already know what params to set for config already. But since we're also moving forward with exposing some of the internal paramters for testing/validation, it's necessary to address this bug.

### Steps to reproduce the behavior

```python
from akd.agents._base import LiteLLMInstructorBaseAgent, BaseAgentConfig

class SomeAgent(LiteLLMInstructorBaseAgent):
...
...

cfg = BaseAgentConfig(...)
agent = SomeAgent(config=cfg)

print(agent.config.temperature) -> Default to 0.0
print(agent.temperature) -> default to 0.0

agent.temperature = 0.5
print(agent.config.temperature) -> 0.0
print(agent.temperature) -> 0.5
```

### Proposed solution

I am working on a bugfix where we dont' need to change anything. rather need to update __set_attrs_from_config() method. Instead of doing setattr(...) we need to create `@property`  with setter and getter (fset/fget)

```python
def __set_attrs_from_config(self):
        if self.config is None:
            return

        # Create properties that reference self.config.field_name instead of copying values
        # This ensures agent.field_name and agent.config.field_name reference the same value
        for field_name in type(self.config).model_fields.keys():
            # Skip if already a property (from exposed params or previous instance)
            if isinstance(getattr(type(self), field_name, None), property):
                continue

            # Create property that references self.config.field_name
            def make_config_property(fname):
                """Create a property that references a config field."""
                def getter(self):
                    return getattr(self.config, fname)
                def setter(self, value):
                    setattr(self.config, fname, value)
                return property(getter, setter)

            # Set property on the class (affects all instances)
            setattr(type(self), field_name, make_config_property(field_name))

        # Also create properties for computed fields (read-only)
        # These are in model_computed_fields, not model_fields
        # Pydantic 2.x supports model_computed_fields
        if hasattr(type(self.config), "model_computed_fields"):
            for field_name in type(self.config).model_computed_fields.keys():
                # Skip if already a property
                if isinstance(getattr(type(self), field_name, None), property):
                    continue

                # Create read-only property for computed fields
                def make_computed_property(fname):
                    """Create a read-only property for computed config field."""
                    def getter(self):
                        return getattr(self.config, fname)
                    return property(getter)

                setattr(type(self), field_name, make_computed_property(field_name))
```

After the bugfix we will have something like this:

Image

NISH1001:
### **Summary** :memo:

This PR resolves \#286 by refactoring how `agent` and `tool` attributes synchronize with their underlying `config` objects.

It moves the property generation logic from instance initialization (`__set_attrs_from_config`) to class definition time using a Metaclass (`AbstractBaseMeta`). This ensures that `agent.field` and `agent.config.field` are always in sync by utilizing Python descriptors.

### **Key Changes** :hammer\_and\_wrench:

1.  **Metaclass Implementation (`AbstractBaseMeta`):**

      * Introduced `_create_config_properties` in the metaclass.
      * Iterates over the `config_schema` fields at class creation time.
      * Dynamically adds `property` descriptors to the class that proxy reads/writes directly to `self.config`.

2.  **Property Factory Functions:**

      * `_make_config_property`: Creates getters/setters for standard fields (bidirectional sync).
      * `_make_computed_property`: Creates getters for `model_computed_fields` (read-only).

3.  **Cleanup:**

      * Removed the legacy manual copying logic in `_post_init` and `__set_attrs_from_config`.
      * Updated `UnrestrictedAbstractBase` to use the new Metaclass.

4.  **Testing:**

      * Added `tests/base/test_config_reference.py` to verify reference semantics for various data types (int, float, bool) and multiple instances.

### **Motivation** :bulb:

Previously, config values were copied to the instance during initialization. This caused two issues:

1.  **Data Desynchronization:** Modifying `agent.temperature` did not update `agent.config.temperature`, leading to inconsistent behavior depending on which attribute was accessed internally.
2.  **Primitive Copying:** Primitive types (int, str, bool) were passed by value, breaking the link between the agent and its config.

By using properties, we ensure a **single source of truth** residing in `self.config`.

### **Usage Example**

The API remains the same, but the behavior is now consistent:

```python
# Assuming HaikuAgent is defined with a config_schema containing 'model_name'

agent = HaikuAgent()

# 1. Check initial synchronization
print(f"Config: {agent.config.model_name}")  # e.g., "gpt-4o-mini"
print(f"Agent:  {agent.model_name}")         # e.g., "gpt-4o-mini"

# 2. Modify via the Agent attribute
agent.model_name = "gpt-4o"

# 3. Verify the Config object automatically updated
assert agent.config.model_name == "gpt-4o"
print("Synchronization successful.")
```

### **Test Coverage** :test\_tube:

  * [x] **Bi-directional updates:** Verifies that setting `agent.x` updates `config.x` and vice versa.
  * [x] **Type integrity:** Tested with `int`, `float`, `str`, and `bool`.
  * [x] **Computed fields:** Verified that `@computed_field` properties remain read-only on the agent.
  * [x] **Instance isolation:** Verified that two instances of the same agent class maintain separate configs.

### **Checklist**

  - [x] Linting passed
  - [x] Tests added (`tests/base/test_config_reference.py`)
  - [x] Backward compatibility maintained (public API is unchanged)

-----

### **What improved:**

1.  **Technical Accuracy:** Correctly identifies that you are now using `AbstractBaseMeta` and `__new__` rather than just `__set_attrs_from_config`.
2.  **Clarity:** Split the "Changes" into specific technical components (Metaclass vs Factory functions).
3.  **Context:** Added the "Instance isolation" point to the test coverage, as I saw you included `test_multiple_agent_instances_have_independent_configs` in the diff.


NISH1001:
PS:
After playing around with different design, gemini also agrees about `@property`:

```
Property Descriptors (Your Current Way)
Targeted: The property descriptor only exists for the specific names you care about (e.g., temperature).
Fast Dispatch: When you access agent.temperature, Python sees the descriptor in the class dictionary and jumps straight to your getter/setter function.
Zero Overhead for Others: When you access agent.internal_state (which is not a config field), it behaves like a normal attribute access. There is zero performance penalty for non-config attributes.

```

muthukumaranR (approved):

muthukumaranR:
Thanks for this solid PR. LGTM - you can add discussed documentations later after - we can merge this.

NISH1001:
TODO:
- I am plannig nto provide more docstring-level documentation for config override etc.

PR #292, opened 2025-12-08
Fix Deepcopy for Agents and Guardrails Attribute Conflicts

NISH1001:
### Summary :memo:
This PR resolves issues with deep-copying agent instances by manually re-initializing API clients and fixes an attribute naming collision in the Guardrails wrapper that was overwriting component fields.

### Details
1. **Agent Deepcopy Support:** Implemented `__deepcopy__` in `InstructorBaseAgent` and `LiteLLMInstructorBaseAgent`. The `openai` and `httpx` clients contain async state/locks that cannot be pickled. The new implementation copies the agent's attributes but re-initializes the `client` fresh.
2. **Guardrails Field Isolation:** Renamed `input_fields` and `output_fields` to `guarded_input_fields` and `guarded_output_fields` within `akd/guardrails.py`. This prevents the Guardrails wrapper from accidentally overwriting existing attributes on the base component.
3. **Guardrails Initialization:** Updated `apply_guardrails` to set these specific fields manually rather than calling `__init__` (which was causing state reset issues) and updated `_extract_text_content` to use the new field names.
4. **Test Updates:** Updated `tests/guardrails/test_field_prioritization.py` to verify the new `guarded_*` attribute names.

### Bugfixes :bug:
- Fixed `TypeError` when attempting to `deepcopy` an agent due to unpickleable client objects.
- Fixed an issue where `apply_guardrails` would conflict with or overwrite a component's native `input_fields`.

### Checks
- [x] Tested Changes
- [ ] Stakeholder Approval

### DIFF
--- a/akd/agents/_base.py
+++ b/akd/agents/_base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 from abc import abstractmethod
 from typing import Any, cast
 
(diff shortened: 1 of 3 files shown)


muthukumaranR (approved):
Thanks for the fix. lgtm

PR #345, opened 2026-02-11
fix: Added unique node IDs, validator for worflow_plan, and unified planner model config

rohit-sahoo:
## Summary
Added unique node IDs to handle duplicate agents, added validator to ensure workflow_plan exists when ready_to_generate is true, and moved planner model config to single source in CONFIG for the planner.

### 1. Unique Node IDs (Duplicate Agents)
- Added `id` field to `WorkflowNode` (e.g., `code_search_1`, `code_search_2`)
- Edges now reference node `id` instead of `type`
- io_map JSONPath uses node `id` (e.g., `$.code_search_1.outputs.results`)
- Updated validation methods for node IDs

### 2. Model Config - Planner
- Added `planner_model_name` to `ModelConfigSettings` as single source of truth
- `PlannerConfig` now uses `CONFIG.planner_model_name`
- `FieldMappingGenerator` now uses `CONFIG.planner_model_name`
- Removed hardcoded model names from planner module

### 3. Validators
- Added validator: `ready_to_generate=True` requires `workflow_plan` to exist
- Prevents crash when LLM says "ready" but forgets to provide plan

## How to Test

**Unit tests:**
```bash
uv run pytest tests/planner/
```

**Manual test:**
```bash
python scripts/demo_planner.py automated "Find weather repos, find climate repos, and do gap analysis"
```

### DIFF
--- a/akd/planner/format_builder.py
+++ b/akd/planner/format_builder.py
@@ -60,12 +60,13 @@ class WorkflowNodeIO(BaseModel):
 class WorkflowNode(BaseModel):
     """Individual node in a workflow definition."""
 
-    type: str = Field(..., description="Node type corresponding to agent type")
+    id: str = Field(..., description="Unique node identifier (e.g., code_search_0, gap_analysis_0)")
+    type: str = Field(..., description="Node type corresponding to agent type for registry lookup")
     input: WorkflowNodeIO = Field(default_factory=WorkflowNodeIO)
     output: WorkflowNodeIO | None = Field(default=None)
     io_map: dict[str, str] | None = Field(
         default=None,
-        description="JSONPath mappings for cross-node data access (target_field: jsonpath_expr)",
+        description="JSONPath mappings for cross-node data access (target_field: jsonpath_expr using node id)",
     )
 
 
@@ -119,7 +120,7 @@ class WorkflowFormat(BaseModel):
 
         Checks that:
         1. JSONPath expressions are well-formed (start with $.)
-        2. Referenced nodes exist in the workflow
+        2. Referenced node IDs exist in the workflow
         3. No circular dependencies exist
 
         Args:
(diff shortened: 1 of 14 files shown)


akd/planner/format_builder.py
```diff
@@ -60,12 +60,13 @@ def model_dump(self, **kwargs) -> dict[str, list]:
 class WorkflowNode(BaseModel):
     """Individual node in a workflow definition."""
 
-    type: str = Field(..., description="Node type corresponding to agent type")
+    id: str = Field(..., description="Unique node identifier (e.g., code_search_0, gap_analysis_0)")
+    type: str = Field(..., description="Node type corresponding to agent type for registry lookup")
```
NISH1001: Since `type` is built-in identifier, maybe switch to something like `type_`?

akd/planner/llm_planner.py
```diff
@@ -75,7 +75,7 @@ class PlannerResponse(OutputSchema):
 
     message: str = Field(..., description="Response message to the user")
     phase: ConversationPhase = Field(..., description="Current conversation phase")
-    question: Optional[PlannerQuestion] = Field(default=None, description="Follow-up question if needed")
+    question: PlannerQuestion | None = Field(default=None, description="Follow-up question if needed")
     workflow_plan: Optional[WorkflowPlan] = Field(default=None, description="Generated workflow plan")
```
NISH1001: can you apply the new type hint style across all the planner modules? wehrever there's older style

NISH1001 (changes requested):

akd/planner/format_builder.py
rohit-sahoo: Updated!

akd/planner/llm_planner.py
rohit-sahoo: Updated

NISH1001 (approved):

PR #354, opened 2026-02-25
fix: deterministic workflow generation, schema enrichment, and enum support for planner

rohit-sahoo:
## Summary
Replaces unreliable LLM-controlled `ready_to_generate` flag with deterministic check, adds schema enrichment to planner input, and fixes enum extraction in agent registry.

## Changes

**1. Deterministic `ready_to_generate` (replaces LLM flag)**
- Removed `auto_set_ready_when_plan_exists` validator from `PlannerResponse`
- Added `is_ready_to_generate()` method to `InteractivePlannerSession`
- Readiness computed from structural conditions: plan exists, has agents, has research goal, all agents in registry
- LLM no longer decides boolean flags, only converses and builds workflow plan

**2. Schema Enrichment (`enrich_workflow_schema`)**
- New `akd/planner/enrich_schema.py` reads from `agent_registry.json` (single source of truth)
- Enriches flat input fields into rich schema objects with type, description, required, default, allowed_values

**3. Enum/Literal Extraction in Registry**
- Added `allowed_values` field to `FieldDefinition`
- Enhanced `_extract_schema()` to detect enum/Literal values directly from Pydantic `model_fields`
- `search_mode` now correctly extracted as `type: "enum"` with `allowed_values: ["fast", "medium", "long", "extensive"]`

**4. Edge Serialization Fix**
- Removed aliases from `WorkflowEdge` updating the outputs `from_node`/`to_node` (backend expected format)

**5. Enum Case Normalization Fix**

Issue:
- LLMs sometimes returns uppercase enum values like "GOAL_CLARIFICATION" instead of "goal_clarification". 
- `_create_instructor_compatible_model()` builds a proxy model with `__base__=BaseModel`, which strips all class-level validators (`@field_validator`, `@model_validator`).
- Both the @field_validator and @model_validator to lowercase these doesn't work Instructor uses a proxy model that strips all validators.

Fix:
- Added `_missing_()` to ConversationPhase and PlannerQuestionType enums so they handle uppercase values themselves. This works because `_missing_` is part of the enum class, not the model, so it can't be stripped. 
- Reverted @model_validator back to the original @field_validator.

**Bug flagged for 5th fix: Enum Case Normalization Fix**

> encountered an error while processing your request: <failed_attempts>\n\n<generation number=\"1\">\n<exception>\n    1 validation error for PlannerResponse\nphase\n  Input should be 'initial_requirements', 'goal_clarification', 'agent_selection', 'io_specification', 'workflow_construction', 'validation' or 'finalization' [type=enum, input_value='GOAL_CLARIFICATION', input_type=str]\n    For further information visit https://errors.pydantic.dev/2.12/v/enum\n</exception>\n<completion>\n    ModelResponse(id='chatcmpl-DCEkVod1CYS3qb9UcBcWab8Tadw2k', created=1771809003, model='gpt-4o-mini-2024-07-18', object='chat.completion', system_fingerprint='fp_373a14eb6f', choices=[Choices(finish_reason='tool_calls', index=0, message=Message(content=None, role='assistant', tool_calls=[ChatCompletionMessageToolCall(function=Function(arguments='{\"message\":\"I\\'ll create a workflow to search for recent research papers on DNA matching methodologies. This will help gather relevant literature on the topic.\",\"phase\":\"GOAL_CLARIFICATION\"}', name='PlannerResponse'), id='call_VjXA4PRjwU9QkLFuXsm8TOWO', type='function')], function_call=None, provider_specific_fields={'refusal': None}, annotations=[]), provider_specific_fields={})], usage=CompletionUsage(completion_tokens=107, prompt_tokens=8200, total_tokens=8307, completion_tokens_details=CompletionTokensDetails(accepted_prediction_tokens=None, audio_tokens=0, reasoning_tokens=0, rejected_prediction_tokens=None), prompt_tokens_details=PromptTokensDetails(audio_tokens=0, cached_tokens=6528)), service_tier='default')\n</completion>\n</generation>\n\n<generation number=\"2\">\n<exception>\n    1 validation error for PlannerResponse\nphase\n  Input should be 'initial_requirements', 'goal_clarification', 'agent_selection', 'io_specification', 'workflow_construction', 'validation' or 'finalization' [type=enum, input_value='GOAL_CLARIFICATION', input_type=str]\n

## How to Test

**Unit tests:**
```bash
uv run pytest tests/planner/ -v
```

**Manual test:**
```bash
rm akd/mapping/agent_registry.json  # regenerate fresh
python scripts/demo_planner.py interactive
```


akd/planner/format_builder.py
```diff
@@ -78,13 +78,8 @@ class WorkflowNode(BaseModel):
 class WorkflowEdge(BaseModel):
     """Edge connecting two nodes in a workflow."""
 
-    from_node: str = Field(..., alias="from", description="Source node ID")
-    to_node: str = Field(..., alias="to", description="Target node ID")
```
NISH1001, line 82: any reason we removed the alias? Was it causing some issues?

akd/planner/llm_planner.py
```diff
@@ -58,13 +58,14 @@ class PlannerQuestion(OutputSchema):
     context: str = Field(..., description="Context explaining why this question is important")
     suggested_answer: str | None = Field(default=None, description="Suggested answer if applicable")
 
-    @field_validator("question_type", mode="before")
+    @model_validator(mode="before")
```
NISH1001: Any reason we switched from field to model validator?

akd/planner/format_builder.py
rohit-sahoo: `serialize_by_alias=True` on the shared model_config was serializing edges as "from" and "to" instead of "from_node" and "to_node" which backend was expecting. So, removed aliases from WorkflowEdge. But the  WorkflowNode keeps its own model_config for the `type_ `-> `type` alias

akd/planner/llm_planner.py
rohit-sahoo: In some situations, `LiteLLMInstructorBaseAgent` which uses the Instructor bypasses Pydantic field_validator casuing the PlannerResponse not getting validated whereas the model_validator is working consistently by both Pydantic and Instructor

NISH1001: Okay. I still need to understand in what scenarios. Is this after the llm generates the json and converting to pydantic object? or even before that process?

rohit-sahoo: I have added the scenario in the PR Summary **5th point: Enum Case Normalization Fix.** 

Even the model_validator was not the correct fix, as this issue can still persist. What happens is the LLM returns value "GOAL_CLARIFICATION" instead of "goal_clarification" and then the `_create_instructor_compatible_model()` builds a proxy model with `__base__=BaseModel`, which strips the @field_validator or @model_validator, so the lowercase is not getting reflected. Reverted back to field_validator and also added a `_missing_()`

NISH1001 (approved):

PR #364, opened 2026-03-04
Wrap astream method with guardrail

paridhi-parajuli:
### Summary :memo:
Add Guardrail Support to Streaming (astream)

### Details
This PR extends the @guardrail decorator to support streaming execution via astream since the agents now use the astream method in the backend.

- Add guardrail enforcement to astream, running input checks before streaming begins.
- Intercept CompletedEvent to apply output guardrails and wrap the final streamed response with the guardrails result


rohit-sahoo (approved):
Looks good to me! 🚢  guardrails hooks into astream and CompletedEvent output wrapped as expected

PR #366, opened 2026-03-05
Fix: enrich input schema for akd-ext based agents

rohit-sahoo:
## Bug:

When akd-ext registers agents at startup via registry.register_agent(CMRCareAgent, persist=False), they exist in the singleton AgentRegistry in memory but not in agent_registry.json. 

### Fix:

1. `enrich_workflow_schema()` was reading agent_registry.json from disk to get input schemas, which missed agents registered at runtime with persist=False (e.g., CMRCareAgent from akd-ext). Now accepts an optional registry parameter and reads schemas from the live in-memory AgentRegistry when provided, falling back to the JSON file when not.
2. `WorkflowBuilder.build()` passes its existing self.registry through to enrich_workflow_schema(), so schema enrichment is consistent with the rest of the build pipeline.

### Changes made:
1. akd/planner/enrich_schema.py:
- Added _build_input_schemas_from_registry() to read schemas from the live registry object. 
- Renamed old _load_input_schemas() → _load_input_schemas_from_file() as fallback. 
- Updated enrich_workflow_schema() to accept optional registry kwarg.

2. akd/planner/workflow_builder.py
- Pass self.registry to enrich_workflow_schema() in build().


akd/planner/enrich_schema.py
```diff
@@ -3,26 +3,60 @@
...
 
-def _load_input_schemas() -> dict[str, dict[str, Any]]:
-    """Build a {agent_id: {field_name: field_schema}} lookup from the registry."""
+def _build_input_schemas_from_registry(registry: AgentRegistry) -> dict[str, dict[str, Any]]:
+    """Build {agent_id: {field_name: field_schema}} lookup from live registry object.
+
+    Uses the in-memory registry which includes both auto-discovered agents
+    and runtime-registered agents (e.g., CMRCareAgent from akd-ext).
+    """
+    schemas: dict[str, dict[str, Any]] = {}
```
muthukumaranR, line 34: Have we run tests on these? e.g. cases where If someone adds a new field attribute later and only updates one path? and other cases where there can be divergence?

muthukumaranR (approved):
lgtm. thanks for the update. just a comment on test cases.

akd/planner/enrich_schema.py
rohit-sahoo: Sure! will add in the next PR with the planner prompt fixes!

PR #418, opened 2026-04-13
enhance: add per-category score to MultiRiskGraniteGuardianTool via logprobs

NISH1001:
### What

- Added `score_threshold` config field to `MultiRiskGraniteGuardianToolConfig` to optionally drop low-confidence detections and reduce false positives. Defaults to `0.0` (no filter) to preserve current behavior.
- Each entry in `risk_results` now includes a per-category `score` in [0, 1], derived from the logprob of the category's first emitted token in Step 2.
- Categories whose per-category score is below `score_threshold` are dropped from `detected_risks` and `risk_results`. Categories without a score (e.g., when Ollama does not return logprobs) pass through unfiltered as a graceful fallback.
- Step 1 decision remains label-based (`Yes`/`No`) as per the model card; no extra Ollama calls are added.

### Why

The multi-harm model's self-reported text confidence (e.g., `"High"`, `"Not Harmful"`) is effectively binary in practice and useless for thresholding. False positives on some categories (e.g., `Harmful` flagged on sarcasm) could not be filtered without manual category exclusion. Per-category logprob-derived scores give callers a real numeric signal to threshold on (e.g., `Violence=0.97` vs `Harmful=0.35` for the same input).

### Scoring methodology (first-token logprob)

We use the **first token's logprob** as the per-category confidence score. Here's how it works with an example:

- Step 2 model output: `Violence, Unethical Behavior, Harmful`
- Ollama returns per-token logprobs for the generated text:

  | Token | logprob |
  |-------|---------|
  | `Violence` | -0.02 |
  | `,` | -0.01 |
  | ` ` | -0.00 |
  | `Un` | -0.70 |
  | `eth` | -0.03 |
  | `ical` | -0.01 |
  | ` Behavior` | -0.02 |
  | `,` | -0.01 |
  | ` ` | -0.00 |
  | `Harm` | -1.20 |
  | `ful` | -0.01 |

- We walk the token stream, skip commas/whitespace, and grab the **first substantive token per category**
- Score = `exp(first_token_logprob)`:
  - **Violence** → `exp(-0.02)` = **0.98** (1 token, first token = full category)
  - **Unethical Behavior** → `exp(-0.70)` = **0.50** (first token = `Un`)
  - **Harmful** → `exp(-1.20)` = **0.30** (first token = `Harm`)

**Why first-token only (not joint probability across all tokens)?**

- The first token is the model's **decision point** — when it emits `Un`, it has already committed to `Unethical Behavior`. Subsequent tokens (`eth`, `ical`, ` Behavior`) are near-deterministic completions (logprobs ~0) that reflect spelling ability, not category confidence.
- **Joint probability** (product of all token probs) introduces **length bias**: longer category names always score lower regardless of model confidence. e.g., `Violence` (1 token) would always outscore `Unethical Behavior` (4 tokens) even if the model is equally confident in both.
- **Geometric mean** (Nth root of joint prob) corrects length bias — maybe worth exploring, but likely not worth the complexity since the first token already captures the model's category selection confidence and subsequent tokens are near-deterministic completions.

### How

- Changes are in `akd/guardrails/providers/granite_guardian.py`
- Added `score_threshold` config (0.0–1.0, default 0.0) to optionally filter low-confidence categories
- Step 2 Ollama request now includes `logprobs: True` and `top_logprobs: 5`
- Category parsing returns a `{category: score}` dict; score filtering and `risk_results` building happen in `_arun`

### How to test

- `uv run pytest tests/guardrails/` — existing tests still pass (8 failures in `test_granite_think.py` are pre-existing and require a live `granite3.3-guardian:8b` model).
- `uv run python scripts/test_multi_harm.py` with live Ollama + `granite-guardian-3.2-5b-multi-harm-GGUF` — verifies per-category scores appear in `risk_results` (e.g., Violence=0.97, Unethical Behavior=0.76 for the same violent input; Harmful=0.35 on sarcasm, correctly flagged as low-confidence).

### DIFF
--- a/akd/guardrails/providers/granite_guardian.py
+++ b/akd/guardrails/providers/granite_guardian.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import math
 import os
 import re
 from enum import StrEnum
@@ -321,6 +322,17 @@ class MultiRiskGraniteGuardianToolConfig(GraniteGuardianBaseConfig):
         ],
         description="Harm categories to report (filters model output, defaults to all harmful).",
     )
+    score_threshold: float = Field(
+        default=0.0,
+        description=(
+            "Minimum per-category score required to include a detected risk. "
+            "Range 0.0-1.0. Default 0.0 = no filtering (all detected categories pass). "
+            "Set (e.g., 0.5) to drop low-confidence detections and reduce false positives. "
+            "Has no effect if Ollama does not return logprobs (scores are None)."
+        ),
+        ge=0.0,
+        le=1.0,
+    )
 
     @model_validator(mode="after")
     def validate_multi_risk_model(self) -> Self:
@@ -384,12 +396,13 @@ class MultiRiskGraniteGuardianTool(GraniteGuardianTool):
         # Step 1: Detect if harmful (Yes/No + confidence)
         step1_result = await self._call_harm_detection(step1_prompt)
 
-        is_harmful = step1_result.get("label", "").lower() == "yes"
+        label = step1_result.get("label", "").lower()
         confidence = step1_result.get("confidence", "")
+        is_harmful = label == "yes"
 
         if self.debug:
             logger.debug(
-                f"[{self.__class__.__name__}] Step 1 - is_harmful={is_harmful}, confidence={confidence}",
+                f"[{self.__class__.__name__}] Step 1 - is_harmful={is_harmful}, label={label}, confidence={confidence}",
             )
 
         if not is_harmful:
@@ -399,7 +412,7 @@ class MultiRiskGraniteGuardianTool(GraniteGuardianTool):
                 risk_results={},
                 provider=self.__class__.__name__,
                 extra={
-                    "risk_label": "no",
+                    "risk_label": label or "no",
                     "confidence": confidence,
                     "step1_raw": step1_result.get("raw_response"),
                 },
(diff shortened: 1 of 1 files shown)


akd/guardrails/providers/granite_guardian.py
```diff
@@ -521,42 +548,101 @@ async def _call_category_detection(self, prompt: str) -> dict[str, Any]:
...
             if self.debug:
                 logger.debug(
                     f"[{self.__class__.__name__}] Step 2 - Ollama response:\n"
                     f"--- RESPONSE START ---\n{content}\n--- RESPONSE END ---",
                 )
 
-            categories = self._parse_categories(content)
+            categories_with_scores = self._parse_categories_with_scores(content, token_logprobs)
+            categories = [cat for cat, _ in categories_with_scores]
+            scores = {cat: s for cat, s in categories_with_scores}
```
NISH1001: Need to refactor this with probably just dict(...)

NISH1001: Done.

akd/guardrails/providers/granite_guardian.py
```diff
@@ -321,6 +322,17 @@ class MultiRiskGraniteGuardianToolConfig(GraniteGuardianBaseConfig):
...
+        default=0.0,
+        description=(
+            "Minimum per-category score required to include a detected risk. "
+            "Range 0.0-1.0. Default 0.0 = no filtering (all detected categories pass). "
+            "Set (e.g., 0.5) to drop low-confidence detections and reduce false positives. "
+            "Has no effect if Ollama does not return logprobs (scores are None)."
+        ),
+        ge=0.0,
+        le=1.0,
+    )
```
muthukumaranR, line 335: should default be higher here? like 0.5 or 0.8 based on how it behaves?

muthukumaranR (commented):
I think we need to figure out if there is a MultiHarm finetuning pipeline so we are not tied to 3.2. it's already outdated by a few versions. 


cc: @jbrry

akd/guardrails/providers/granite_guardian.py
NISH1001: It's for graceful filter. I don't want to have it be destructive. So, with 0.0, it still maintains the original behavior and then we can on backend side control and test threshold

NISH1001:
> I think we need to figure out if there is a MultiHarm finetuning pipeline so we are not tied to 3.2. it's already outdated by a few versions.
> 
> cc: @jbrry

True. We should probalby figure out that soon with LORA

muthukumaranR (approved):

PR #420, opened 2026-04-14
Release: AKD v0.1.1

NISH1001:
## Summary

Patch release v0.1.1 — adds per-category logprob scoring to the multi-risk Granite Guardian tool and consolidates version management.

- **Per-category logprob scores**: `MultiRiskGraniteGuardianTool` Step 2 now requests token logprobs from Ollama and derives a `score = exp(first_token_logprob)` per detected harm category, giving callers a real numeric confidence signal (e.g., `Violence=0.97` vs `Harmful=0.35`)
- **Score threshold filtering**: New `score_threshold` config field (0.0–1.0, default 0.0) to optionally drop low-confidence category detections and reduce false positives
- **Single version source**: Replaced hardcoded `__version__` in `akd/__init__.py` with `importlib.metadata.version("akd")` — version is now sourced solely from `pyproject.toml`

### Changes since v0.1.0
- `abd39dd` Consolidate _parse_categories into single dict-returning function
- `0ea08ee` Add per-category score to MultiRiskGraniteGuardianTool via Step 2 logprobs
- `7fc0863` Bump version to 0.1.1, use importlib.metadata for single version source

## Test plan
- [ ] Verify CI passes on this PR
- [ ] Confirm all existing tests pass (`uv run pytest`)
- [x] `uv run python -c "import akd; print(akd.__version__)"` prints `0.1.1`

### DIFF
--- a/akd/__init__.py
+++ b/akd/__init__.py
@@ -11,7 +11,9 @@ research through:
 - Framework agnostic core logic decoupled from orchestration engines
 """
 
-__version__ = "0.1.0"
+from importlib.metadata import version as _version
+
+__version__ = _version("akd")
 __author__ = "NASA IMPACT"
 __email__ = "np0069@uah.edu,mr0051@uah.edu"
 
--- a/akd/__init__.py
+++ b/akd/__init__.py
@@ -11,9 +11,13 @@ research through:
 - Framework agnostic core logic decoupled from orchestration engines
 """
 
+from importlib.metadata import PackageNotFoundError
 from importlib.metadata import version as _version
 
-__version__ = _version("akd")
+try:
+    __version__ = _version("akd")
+except PackageNotFoundError:
+    __version__ = "0.1.1"
 __author__ = "NASA IMPACT"
 __email__ = "np0069@uah.edu,mr0051@uah.edu"
 
--- a/akd/guardrails/providers/granite_guardian.py
+++ b/akd/guardrails/providers/granite_guardian.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import math
 import os
 import re
 from enum import StrEnum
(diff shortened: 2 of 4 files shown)


sanzog03 (approved):

akd/__init__.py
```diff
@@ -11,7 +11,9 @@
 - Framework agnostic core logic decoupled from orchestration engines
 """
 
-__version__ = "0.1.0"
+from importlib.metadata import version as _version
+
+__version__ = _version("akd")
```
sanzog03: will this work when the code is directly used from source code without installing it?

akd/guardrails/providers/granite_guardian.py
```diff
@@ -542,21 +570,69 @@ async def _call_category_detection(self, prompt: str) -> dict[str, Any]:
...
+        for i, raw_cat in enumerate(raw_categories):
             try:
-                categories.append(GraniteHarmCategory(raw_cat))
+                cat = GraniteHarmCategory(raw_cat)
             except ValueError:
-                logger.warning(f"[MultiHarmGraniteGuardianTool] Unknown category: {raw_cat}")
+                logger.warning(f"[{self.__class__.__name__}] Unknown category: {raw_cat}")
+                continue
+            lp = first_token_logprobs[i] if i < len(first_token_logprobs) else None
+            results[cat] = math.exp(lp) if lp is not None else None
```
sanzog03, line 604: can lp be too large? if too large, could cause overflow error.

akd/__init__.py
NISH1001: It will not work. So, reocmmended way is to always use uv sync to install since we always ask to use uv sync. Probalby need to catch error and fallback to some version, the latest one

akd/guardrails/providers/granite_guardian.py
NISH1001: I don't think it will happen since they are log probs from ollama and they should probably always be negative or 0 --> exp(logprob) would lie in 0, 1 range. Unless ollama behaves abnormally.

akd/__init__.py
NISH1001: I have adde some fallback guard for this. Thanks for suggestion.

PR #431, opened 2026-04-20
Add is_empty structural emptiness check for OutputSchema

NISH1001:
## Summary

- Add `akd.utils.is_empty(value)` — recursive structural emptiness check. Returns `True` for `None`, empty/whitespace strings, empty bytes, and containers (`BaseModel`, `dict`, `list`, `tuple`, `set`, `frozenset`) whose contents are all recursively empty. Concrete scalars like `0`, `False`, `0.0`, `datetime(...)` are NOT considered empty.
- Expose `OutputSchema.is_empty()` as a thin wrapper so callers can do `result.is_empty()` instead of hand-rolling per-schema checks. Useful for short-circuiting downstream logic when an agent/tool returns a structurally blank result.
- Lazy-import of `akd.utils` inside the method to avoid a circular import via `akd/__init__.py`.

## Test plan

- [x] `uv run pytest tests/test_utils_is_empty.py -v` — 43 passed
- [x] Manual sanity: `O().is_empty()` True; `O(a='hi').is_empty()` False; `O(a='   ').is_empty()` True
- [x] `uv run pytest` — full suite green

### DIFF
--- a/akd/_base/_base.py
+++ b/akd/_base/_base.py
@@ -110,6 +110,12 @@ class OutputSchema(IOSchema):
         """Get the run context associated with this output."""
         return self._run_context
 
+    def is_empty(self) -> bool:
+        """Structurally empty if all fields are None/empty/empty-of-empty (recursive)."""
+        from akd.utils import is_empty as _is_empty
+
+        return _is_empty(self)
+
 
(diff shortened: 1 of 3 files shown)


sanzog03 (approved):
lgtm

